PR #48, opened 2017-02-09
added new OPLS types to XML file

summeraz:


summeraz:
Alright, all of the atom types that were added to the OPLS force field XML file should now be included in small molecule atom typing tests.

The formic acid test appears to be failing now, which is a bit confusing to me, as it is finding two possible atom types where it should only be left with one... This does, however, bring up a good point about which rule should be chosen if an atom is a part of multiple functional groups (in the case of formic acid should aldehyde or carboxylic acid rules be used?).  There are a lot of combinations of functional groups to consider, but eventually some additional overrides will need to be included for those cases.

PR #58, opened 2017-02-14
Tidy up OPLS SMARTS

ctk3b:


davidlmobley:
Perhaps we should touch bases again soon. How are you guys indicating what atom the "atom type" refers to in your SMARTS patterns? Is it the one in square brackets? I'm just trying to wrap my mind around whether making a converter from this format to our format or vise versa would be trivial vs a pain in the neck.

ctk3b:
> Perhaps we should touch bases again soon. 

Agreed. Sorry we fell out of contact!

> How are you guys indicating what atom the "atom type" refers to in your SMARTS patterns? Is it the one in square brackets?

It's simply the first atom. Proper documentation is in the works! 

> I'm just trying to wrap my mind around whether making a converter from this format to our format or vise versa would be trivial vs a pain in the neck.

I've revisited this a few times in the past months and tinkered a bit with RDKit. One of the reasons we fell out of touch in the short term is that we have been unsuccessful in acquiring an OpenEye license (I believe we've applied 5 times now across 3 individuals). That being said, I am 100% certain that if we badgered a few more people that we could get the license but part of what we're trying to do is encourage people to properly disseminate forcefields in a more reproducible manner. To do this, we're trying to keep the barriers to entry as absolutely minimal as possible. 

If you have any more thoughts on how to streamline the licensing issue, we'd love any more information and unless that can be significantly simplified, I'd like to talk more about our options with RDKit. If that is a feasible replacement, I do not think it should be too challenging to convert from a foyer force field to SMIRFF.

davidlmobley:
I just e-mailed; I think the OpenEye license issue can be greatly improved in all likelihood, and I sent some suggestions on how to do it. 

We're also receptive to RDKit, but don't currently have developer time to move things over. If you had someone who could help us with that, we'd be willing to move that direction.

PR #90, opened 2017-03-09
Add OPLS atomtypes for dimethylformadmide

mattwthompson:


ctk3b:
I would just resubmit the whole thing from a fresh branch off of master

PR #94, opened 2017-03-11
Include DOI references in force field XML files

summeraz:
A `doi` attribute can now be included in force field XML files.  The doi references for all atom types in a structure can be printed to a JSON file by providing an additional argument to `apply`.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -23,6 +23,7 @@ requirements:
     - six
     - openmm
     - plyplus
+    - requests
 
 test:
   requires:
--- a/environment.yml
+++ b/environment.yml
@@ -11,6 +11,7 @@ dependencies:
 - networkx
 - six
 - plyplus
+- requests
 # Testing Deps
 - pytest >=3.0
 - python-coveralls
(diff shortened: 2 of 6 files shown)


summeraz:
Right now the file that is generated has a default filename that cannot be changed by the user.  This is probably not a good idea...but I also didn't want to add too many new arguments to `apply`.  Also, I'm not sure JSON is the best format for writing these.

Right now a DOI is printed for each atom type in a structure that has had a force field applied.  I'm not sure if this is the behavior we want, or if we just want to print out the unique DOI references.

I've included some DOI references in the OPLS force field XML file, only those references found in 10.1021/ja9621760.  As it turns out some of the parameters in 10.1021/ja9621760 are duplicates from older OPLS papers (for example, the parameters for opls_145 are included in 10.1021/ja9621760, but were actually derived in 10.1021/ja00168a022), so I'm not sure which DOI is best to use.

ctk3b:
The argument `output_refs` could just accept a filename to write them too. 

I think ideally, the output file should be something like a bibtex file. You can use a snippet like https://gist.github.com/jrsmith3/5513926 to resolve the DOIs to a `.bib` entry.

summeraz:
Something else that could be useful would be having the ability to add DOI's to bonded parameters.  This is a little bit trickier, and I'm not sure how exactly this would work yet since I think currently those are all read by OpenMM.

ctk3b:
I think we'd just need to override their bond force generator to inject an extra step. 

PR looks good though 👍

PR #97, opened 2017-03-15
Graph based SMARTS matching

### LINKED ISSUE #77: Re-enable rule map building to speedup execution
Maps rules to element types so we only run rules for the given element.

### LINKED ISSUE #79: Make atom typing and force field applying separate and optional
Some users may already have an atomtyped system that just needs a force field applied. Others may just want to atom type something without applying the whole force field.

ctk3b:
SMARTS strings are now read into a `SMARTSGraph` object which can be matched against an `OpenMM.Topology` and return all matching atoms. This replaces the functionality in the `Rule` class which operated on one atom at a time.

Also contains fixes that resolve #79 and resolve #77


TengyuMaVandy:
Hi Christoph, is this the improved code of the NetworkX support? Should I close the previous pull request?

ctk3b:
It is. You can probably close it but we won't merge this yet until we can confirm it doesn't cause any performance issues (which I believe it currently does).

PR #98, opened 2017-03-15
Add TraPPE united-atom force field

summeraz:
A force field XML is added for TraPPE united-atom alkanes and alcohols (some alkene parameters are included as well although these are incomplete and commented out).  A library of small molecules has been added for atom typing unit tests.


foyer/tests/utils.py
```diff
@@ -3,8 +3,7 @@
 import numpy as np
 
 
-
-def atomtype(structure, forcefield):
+def atomtype(structure, forcefield, coarse_grained=False):
```
ctk3b: Let's rename this to `non_atomistic`

ctk3b:
I think to fix appveyor change [this line](https://github.com/mosdef-hub/foyer/blob/master/devtools/conda-recipe/meta.yaml#L36) to:

```bash
- foyer/tests/*.txt
```

foyer/tests/utils.py
```diff
@@ -3,8 +3,7 @@
 import numpy as np
 
 
-
-def atomtype(structure, forcefield):
+def atomtype(structure, forcefield, non_atomistic=False):
```
ctk3b: Is this `non_atomistic` flag only required here because this renaming would normally happen in `Forcefield` and doesn't in the way we're doing the tests?

summeraz: Correct.  In `Forcefield` we create new 'elements' for the non-atomistic particles, but the tests are taking elements straight from the MOL2 reader.

foyer/tests/test_forcefield.py
```diff
@@ -23,7 +23,7 @@ def test_load_files():
         ff2 = Forcefield(forcefield_files=ff_file)
         assert len(ff1._atomTypes) == len(ff2._atomTypes)
 
-        ff3 = Forcefield(name=ff_file.split('/')[-1].strip('.xml'))
+        ff3 = Forcefield(name=os.path.split(ff_file)[-1].strip('.xml'))
```
ctk3b: `os.path.splitext(os.path.basename(ff_file))[0]`

PR #103, opened 2017-03-19
Convert OPLS testing files from gro/top to mol2

mattwthompson:
I'm finding mol2 files much easier to work with than pdb, so I used parmed to converted some OPLS test molecules from gro + top to atomtyped mol2.

For reference (surely could be done cleaner)
```python
import parmed as pmd 
import os

cwd = os.getcwd()

for root, subdir, files in os.walk(cwd):
    if any(['top' in f for f in files]):
        gro_file = [x for x in files if 'gro' in x][0]
        top_file = [x for x in files if 'top' in x][0]
        dir_name = gro_file[:-4]
        gro_file = dir_name + '/' + gro_file
        top_file = dir_name + '/' + top_file
        try:
            structure = pmd.load_file(top_file,
                xyz=gro_file)
            structure.save(dir_name + '/' +
                dir_name + '.mol2', overwrite=True)
        except:
            continue
```


ctk3b:
> I'm finding mol2 files much easier to work with than pdb

* Do you mean the `.gro`/`.top` combo?

* There's very odd failure that only appears on Linux Travis which seems to think that there's an oxygen in cyclohexane. I have no idea how this is happening but one issue is that there are still all the `.gro`/`.top` files lying around.

* That being said, I'm a bit hesitant to merge this because I can't really review this without 100% trusting the automated conversion. I agree that working with just `.mol2` files might be easier but what problem specifically does this alleviate since it's just the validation suite?

mattwthompson:
Yes I meant gro/top.

I haven't gotten to the bottom of the cyclohexane issue, but I tested some molecules after deleting top files and they were still passing. Not sure about the specific linux issue, as the atomtypes look the same to me in the file. I can dig more later if necessary.

This definitely doesn't add much to the package, but it's been a annoying when building SMARTS strings to have the topology and atomtypes in one file and the coordinates in different files while trying to visualize it in VMD to better understand the bond network.

Regardless, this should be done more cleanly in a library of molecules somewhere else, something I plan to get to soon. So I'll have my own version of this if it's not added to foyer.

PR #110, opened 2017-04-11
added atom type key reference checks

sallai:
These are  more advanced uses of XML Shemas to make sure that types in nonbonded interaction defintions match the atomtype names.


ctk3b:
Subsumed by #112

Issue #96, opened 2017-03-15
Coarse-grained "element" names cannot contain other coarse-grained element names


summeraz:
If I have three coarse-grained atom types with "elements" `_CH3`, `_CH2`, and `_CH`, and a SMARTS string that looks like `[_CH3;X1]`, the SMARTS parser fails.  The parser will recognize `_CH` as a valid element and then yield an error because it doesn't know what to do with the `3`.  Since `_CH3` is not a valid SMARTS if `_CH` is assigned as the element, there may be some way we can alter the parser such that if an invalid character is detected after an element, the element is reassigned.

Issue #64, opened 2017-02-16
Verification of force field files


ctk3b:
* .xml schema to catch typos in attributes/tags
* run SMARTS parsing on every atomtype to catch syntactic errors
* ensure ampersands only exist in SMARTS strings, then run something like [this](http://stackoverflow.com/questions/4972210/escaping-characters-in-a-xml-file-with-python)

mikemhenry:
For example, overrides != override. I think its fine if we want the force field file but it should at least throw a warning with an non keyword argument is passed.

PR #112, opened 2017-04-12
Add xml based validation tests

ctk3b:
Still needs to be completed and tidied up but it's useful already so let's try to get it merged in.


ctk3b:
Resolves #96 and resolves #64

PR #114, opened 2017-04-13
Added usage examples markdown file

summeraz:


docs/usage_examples.md
```diff
@@ -0,0 +1,88 @@
...
+also allows for separate force fields to be used to atom-type separate portions of a
+molecular system.
+
+#### Creating a box of ethane
+Here we use mBuild to construct a box filled with ethane molecules and use Foyer to
+atom-type the system, applying the OPLS force field, and save to run-able GROMACS 
+files.
+```python
+import mbuild as mb
+from mbuild.examples.ethane.ethane import Ethane
```
ctk3b: I think you can just do `from mbuild.examples import Ethane`. We pulled up a few of the main examples into that namespace.

docs/usage_examples.md
```diff
@@ -0,0 +1,88 @@
...
+
+ethane_box = mb.fill_box(Ethane(), 100, box=[2, 2, 2])
```
ctk3b: Let's add the `n_molecules` kwarg

docs/usage_examples.md
```diff
@@ -0,0 +1,88 @@
...
+ethane_box.save('ethane-box.gro')
+ethane_box.save('ethane-box.top', forcefield_name='oplsaa')
+```
+----
+#### Creating a box of coarse-grained ethane
+Again we will use mBuild to construct a box filled with ethane molecules.  However,
+now we will model ethane using a united-atom description and apply the TraPPE force
+field during atom-typing.  Note how particle names are prefaced with an underscore so
```
ctk3b: prefaced->prefixed

docs/usage_examples.md
```diff
@@ -0,0 +1,88 @@
...
+    def __init__(self):
+        super(CH3_UA, self).__init__()
+        self.add(mb.Particle(name='_CH3'))
+        self.add(mb.Port(anchor=self[0], separation=0.075), 'up')
+
+ethane_UA = mb.Compound()
+ethane_UA.add(CH3_UA(), 'ch3_1')
+ethane_UA.add(CH3_UA(), 'ch3_2')
+mb.force_overlap(ethane_UA['ch3_1'], ethane_UA['ch3_1']['up'],
+                 ethane_UA['ch3_2']['up'])
```
ctk3b: Might be worth adding a CG example to mbuild so we don't need to spend time here explaining how to build this since it's not really pertinent to foyer itself. Or maybe the class could be simplified to something like:

```python
ethane_UA = mb.Compound()
ch3_1 = mb.Particle(name='_CH3', pos=[0, 0, 0])
ch3_2 = mb.Particle(name='_CH3', pos=[0.2, 0, 0])
ethane_UA.add([ch3_1, ch3_2])
ethane_UA.add_bond((ch3_1, ch3_2))
```

docs/usage_examples.md
```diff
@@ -0,0 +1,88 @@
...
+ethane_UA_box.save('ethane-UA-box.top', forcefield_name='trappe-ua')
+```
+----
+#### Combining force fields
+In some instances, the use of multiple force fields may be desired to describe a
+molecular system.  For example, the user may want to use one force field for a
+surface and another for a fluid in the same system.  Foyer supports this
+functionality by allowing the user to separately atom-type parts of a system.  In
+this example, we take a system featuring bulk united atom ethane above a silica 
+surface and apply the OPLS force field to the surface and the TraPPE force field to
```
ctk3b: I think generally this should be discouraged as TraPPE and OPLS aren't really intended to play nicely together. I think the bigger motivation is for when you have a small extension of a larger monolithic forcefield. You could do this exact example with the big OPLS and then just the silica extension from the Sandia papers.

docs/usage_examples.md
```diff
@@ -0,0 +1,88 @@
...
+from foyer.test.utils.import get_fn
+import mbuild as mb
+from mbuild.lib.atoms import H
+from mbuild.lib.bulk_materials import AmorphousSilica
+
+class CH3_UA(mb.Compound):
+    def __init__(self):
+        super(CH3_UA, self).__init__()
+        self.add(mb.Particle(name='_CH3'))
+        self.add(mb.Port(anchor=self[0], separation=0.075), 'up')
```
ctk3b: Related to the above comment, for the AA system you can probably make all this much simpler and maybe just import examples from mBuild instead of spending a lot of time building stuff here.

summeraz:
I've made the suggested edits.  One issue now is that the '+' syntax for combining ParmEd Structures is yielding an error that a maximum recursion depth has been exceeded.  I've attached the script I've been working with, and will continue to attempt to debug this myself.
[multi-ff.zip](https://github.com/mosdef-hub/foyer/files/922123/multi-ff.zip)

PR #115, opened 2017-05-03
Add validation for overrides

ctk3b:


mattwthompson:
Maybe minor but it might be useful for ```doi``` to also be in the bonded classes.

PR #117, opened 2017-05-12
Pass residue information from Parmed Structure to OpenMM Topology

summeraz:
This preserves residue information during atomtyping.  Previously all atoms were lumped into a single residue.


summeraz:
I'll put together a test for this as well

PR #120, opened 2017-05-15
Add 'residues' argument to 'apply'

rmatsum836:


ctk3b:
Thanks @rmatsum836. Assuming this test passes locally, then this is just a matter of waiting for the next mbuild release to happen which includes the `residues` functionality.

rmatsum836:
So I ran this test locally @ctk3b and it passed.  Does your previous comment mean that tests in Travis will fail until the next mbuild release?

ctk3b:
> Does your previous comment mean that tests in Travis will fail until the next mbuild release?

Exactly. When Travis builds foyer and runs the tests, it is installing the latest mbuild release, not building it from source.

PR #123, opened 2017-05-23
Parmed customelement fix

ahy3nz:
Read custom atom types from pdb, correctly create elements based on element names in non_element_types

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,4 +4,4 @@ graft examples
 include foyer/tests/implemented_opls_tests.txt
 graft foyer/trappe_validation
 include foyer/tests/implemented_trappe_tests.txt
-global-include *.xml *.mol2 *.top *.gro
+global-include *.xml *.mol2 *.top *.gro *.pdb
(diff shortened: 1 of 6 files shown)


ctk3b:
This looks good, thank you! Could you also file an issue on mBuild outlining the problem you found with uppercasing the names when starting with an mBuild compound loaded by a .pdb?

One last thing to fix though which is breaking appveyor: add a line here that also includes `*.pdb`

https://github.com/mosdef-hub/foyer/blob/master/setup.py#L90

ctk3b:
Two more. We should probably fix this more elegantly in the near future.

https://github.com/mosdef-hub/foyer/blob/master/MANIFEST.in
https://github.com/mosdef-hub/foyer/blob/master/devtools/conda-recipe/meta.yaml#L33

ahy3nz:
https://github.com/mosdef-hub/foyer/blob/master/devtools/conda-recipe/meta.yaml#L36
Is the meta.yaml file already looking for files in foyer/tests/files?

ctk3b:
Oh yeah sorry, I was thinking it should technically also require the trappe_validation folder but the fact that it isn't causing errors suggests it's unnecessary.

ahy3nz:
It looks all the checks have passed, is there anything else that needs to be done?

ctk3b:
Great, thank you @ahy3nz !

PR #125, opened 2017-06-05
Add Improper in xsd

mattwthompson:
I couldn't find anything special in the docs that's special about impropers. Defined by the same number of parameters and uses the same functional form.

http://docs.openmm.org/7.0.0/userguide/application.html#rbtorsionforce


ctk3b:
Looks like you haven't pulled the latest commits from master. Can you make sure not to delete that work in this PR?

PR #126, opened 2017-06-05
Add Improper in xsd

mattwthompson:
Figured it was easier to start fresh than fix


ctk3b:
Sweet thanks. Will merge after tests pass.

ctk3b:
Is this more or less working? Can you remove the set_trace() from the test?

mattwthompson:
Right now the impropers are implemented for periodic torsion but not RB. The tests pass locally and I've actually been using foyer-generated impropers in one of my systems (at least they're being written all the way to .top).

PR #127, opened 2017-07-05
Changed "class" for atomtypes 151, 152, 153 on 'oplsaa.xml'

rmatsum836:
The "class" on opls_151, opls_152, and opls_153 SMARTS strings were updated.  This should fix missing bond information for alkyl chlorides.

### DIFF
--- a/foyer/version.py
+++ b/foyer/version.py
@@ -1,0 +1,7 @@
+
+# This file is generated in setup.py at build time.
+version = '0.3.1'
+short_version = '0.3.1'
+full_version = '0.3.1'
+git_revision = '1271e10973d4741b3b9a7a0ec3f4c2b3d9ae0052'
+release = True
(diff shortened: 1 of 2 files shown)


foyer/version.py
```diff
@@ -0,0 +1,7 @@
+
```
summeraz: We don't want to merge this file (in fact, version.py should probably be added to .gitignore).

summeraz:
The changes to the OPLS XML look good.  Just need to remove version.py from the PR and we should be good to merge, although I'll need to look a bit more into why the AppVeyor builds are failing...

summeraz:
The Appveyor build for Python 3.4 is failing, but this is because we're now only supporting 2.7 and 3.5, so I'm going to go ahead and merge this PR and then edit the Appveyor.yml file so that we no longer test on 3.4.

PR #133, opened 2017-07-19
Slice out "rules" for elements not found in system

mattwthompson:
Added a pre-processing step in ```find_atomtypes``` that internally throws out rules based on atomtypes not found in the system and reduces the size of the ```rules``` dictionary accordingly. For example, if you were atomtyping alkanes with OPLS, it would remove rules considering N, O, F, etc. before atomtyping. This does not, however, modify the ```Forcefield``` itself because the user may wish to re-use that forcefield for a system that contains a different set of unique elements.

### DIFF
--- a/foyer/atomtyper.py
+++ b/foyer/atomtyper.py
@@ -35,7 +35,7 @@ def find_atomtypes(topology, forcefield, max_iter=10):
                     element = None
         else:
             element = None
-        if element in system_elements:
+        if element is None or element in system_elements:
             subrules[key] = val
     rules = subrules
 
--- a/foyer/atomtyper.py
+++ b/foyer/atomtyper.py
@@ -18,6 +18,27 @@ def find_atomtypes(topology, forcefield, max_iter=10):
 
     """
     rules = _load_rules(forcefield)
+
+    # Only consider rules for elements found in topology
+    subrules = dict()
+    system_elements = {a.element.symbol for a in topology.atoms()}
+    for key,val in rules.items():
+        atom = val.node[0]['atom']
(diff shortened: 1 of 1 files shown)


foyer/atomtyper.py
```diff
@@ -18,6 +18,27 @@ def find_atomtypes(topology, forcefield, max_iter=10):
...
+                element = atom.select('atom_symbol').strees[0].tail[0]
+            except IndexError:
+                try:
+                    atomic_num = atom.select('atomic_num').strees[0].tail[0]
+                    element = pt.Element[int(atomic_num)]
+                except IndexError:
+                    element = None
+        else:
+            element = None
+        if element in system_elements:
```
ctk3b: I believe this still tosses out rules whose element resolves to `None`.

I think either this line should be:

`if element is None or element in system_elements`

or you just add `None` to `system_elements` before the loop

ctk3b:
Excellent, thank you. You tested this on a few systems right? What are the ballpark speedups/system specs?

mattwthompson:
I was messing around with my ILs earlier, but here I did a benchmark of 200 hexane molecules using the existing OPLS force field.

Using the existing master branch, atomtyping takes 55 seconds, and on this branch it takes 48 seconds. For heaxane, it drops the number of rules from 89 to 67. However, if I turn all of the hydrogens into fluorines, it is cut down to 48 rules and the comparison is 30 to 23 seconds. So it may typically be 10-20% but it probably varies with the size of the original force field and the chemistry of the system.

chrisiacovella:
Can you do a quick test as a function of system size (e.g., 2000, 20000 hexanes)?  Does this scale with system-size, as one might expect.

PR #138, opened 2017-08-14
Avoid duplicate residue atomptying

mattwthompson:
Big thanks to @summeraz and @justinGilmer for codemonkey-ing this out.

Currently, each residue in a system is atomtyped even if an identical residue was previously atomtyped. For example, the current atomtyper would atomtype each water in a million-water system, which is a waste of time for larger systems. The approach here is to create a map of unique, already-atomtyped residues, so the 2nd to Nth residue don't call the subgraph isomorphism routines and instead look back at a prototype that was added to the map when it was first found.

* It appears to not be possible to slice ```openmm.app.Topology``` so I wrote a helper function that takes a ```openmm.app.Topology.Residue``` and returns an otherwise-identical ```openmm.app.Topology```. This could be done cleaner if slicing is implemented in the future. This can casually be done in ```ParmEd``` but we found it to be expensive and unnecessary to move between the two types. We tested a number of different combinations of going between the two, but found it was best to stay in openmm the entire time.

* The case of atoms in a residue being bonded to other residues is not trivial to handle, so I added a check for this, and then use the current atomtyping method if this is the case. We might want to revisit this for large polymer or protein systems, but this works fine when so-called "residues" are actually separate molecules.

There is still some overhead associated with this method, but the result is much cheaper for these types of systems. I have done some benchmarking on a system of butane molecules to show this:

image

The "brute-force" curve is from the current master branch and "with-residues" is from this branch. I was unable to run the 100,000 residue systems, even on the Rahman head node, due to memory issues. The speedup converges to a factor of about 20, which I suspect could be improved upon in the future. The linear scaling is a bit of a concern, but I haven't gotten to the bottom of it yet.

### DIFF
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -136,19 +136,24 @@ def _topology_from_residue(res):
         topology_atom.bond_partners = []
 
     for bond in res.bonds():
-        try:
-            atom1 = atoms[bond.atom1]
-        except KeyError:
-            import pdb; pdb.set_trace() 
-        try:
-            atom2 = atoms[bond.atom2]
-        except KeyError:
-            import pdb; pdb.set_trace() 
+        atom1 = atoms[bond.atom1]
+        atom2 = atoms[bond.atom2]
         topology.addBond(atom1, atom2)
         atom1.bond_partners.append(atom2)
         atom2.bond_partners.append(atom1)
 
     return topology
+
+
+def _check_independent_residues(topology):
+    """Check to see if residues will constitute independent graphs"""
+
+    for res in topology.residues():
+        atoms_in_residue = set([atom for atom in res.atoms()])
+        bond_partners_in_residue = [item for sublist in [atom.bond_partners for atom in res.atoms()] for item in sublist]
+        if set(atoms_in_residue) != set(bond_partners_in_residue):
+            return False
+    return True
 
 
 def _update_atomtypes(unatomtyped_topology, atomtyped_prototype_topology):
@@ -340,18 +345,26 @@ class Forcefield(app.ForceField):
             the newly created System
         """
         if atomtype:
-            residue_map = dict()
+            independent_residues = _check_independent_residues(topology)
 
-            for res in topology.residues():
-                if res.name not in residue_map.keys():
-                    residue = _topology_from_residue(res)
-                    find_atomtypes(residue, forcefield=self)
-                    residue_map[res.name] = residue
-        
-            for key, val in residue_map.items():
-                new_system = _update_atomtypes(topology, val)
-        
-            topology = new_system
+            if independent_residues == True:
+                residue_map = dict()
+
+                for res in topology.residues():
+                    if res.name not in residue_map.keys():
+                        residue = _topology_from_residue(res)
+                        find_atomtypes(residue, forcefield=self)
+                        residue_map[res.name] = residue
+
+                new_topology = topology
+
+                for key, val in residue_map.items():
+                    new_topology = _update_atomtypes(topology, val)
+
+                topology = new_topology
+
+            else:
+                find_atomtypes(topology, forcefield=self)
 
         data = app.ForceField._SystemData()
         data.atoms = list(topology.atoms())
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -120,16 +120,27 @@ def _topology_from_parmed(structure, non_element_types):
 
 
 def _topology_from_residue(res):
-    """Converts a openmm.app.Topology.Residue to openmm.app.Topology"""
+    """Converts a openmm.app.Topology.Residue to openmm.app.Topology.
+
+    Parameters
+    ----------
+    res : openmm.app.Topology.Residue
+        An individual residue in an openmm.app.Topology
+
+    Returns
+    -------
+    topology : openmm.app.Topology
+        The generated topology
+
+    """
     topology = app.Topology()
     chain = topology.addChain()
     new_res = topology.addResidue(res.name, chain)
 
-    atoms = dict()  # omm.Atom in res : omm.Atom in topology
+    atoms = dict()  # { omm.Atom in res : omm.Atom in *new* topology }
 
     for res_atom in res.atoms():
         topology_atom = topology.addAtom(name=res_atom.name,
-                         #element=elem.Element.getBySymbol(res_atom.name),
                          element=res_atom.element,
                          residue=new_res)
         atoms[res_atom] = topology_atom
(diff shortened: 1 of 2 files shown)


foyer/forcefield.py
```diff
@@ -340,18 +345,26 @@ def createSystem(self, topology, atomtype=True, nonbondedMethod=NoCutoff,
...
-                    find_atomtypes(residue, forcefield=self)
-                    residue_map[res.name] = residue
-        
-            for key, val in residue_map.items():
-                new_system = _update_atomtypes(topology, val)
-        
-            topology = new_system
+            independent_residues = _check_independent_residues(topology)
+
+            if independent_residues == True:
```
justinGilmer: In general, when checking if a boolean value is True or False, it is better to just check like so:
```python
if independent_residues:
   ...
```

foyer/forcefield.py
```diff
@@ -118,6 +118,53 @@ def _topology_from_parmed(structure, non_element_types):
...
+    """Converts a openmm.app.Topology.Residue to openmm.app.Topology"""
+    topology = app.Topology()
+    chain = topology.addChain()
+    new_res = topology.addResidue(res.name, chain)
+
+    atoms = dict()  # omm.Atom in res : omm.Atom in topology
+
+    for res_atom in res.atoms():
+        topology_atom = topology.addAtom(name=res_atom.name,
+                         #element=elem.Element.getBySymbol(res_atom.name),
```
justinGilmer: stray comment?

foyer/forcefield.py
```diff
@@ -118,6 +118,53 @@ def _topology_from_parmed(structure, non_element_types):
...
     return topology, positions
 
+
+def _topology_from_residue(res):
+    """Converts a openmm.app.Topology.Residue to openmm.app.Topology"""
+    topology = app.Topology()
+    chain = topology.addChain()
+    new_res = topology.addResidue(res.name, chain)
+
+    atoms = dict()  # omm.Atom in res : omm.Atom in topology
```
justinGilmer:
```python
# {omm.Atom from res: omm.Atom in *new* topology}
```
Just to show dictionary relation a bit cleaner.
And this is the newly generated Topology correct?

Maybe add a bit more detail to comment, or a more detailed docstring. Not necessary in terms of PEP8 for private methods, but could be useful for debugging later.

foyer/forcefield.py
```diff
@@ -118,6 +118,53 @@ def _topology_from_parmed(structure, non_element_types):
...
+
+    for res in topology.residues():
+        atoms_in_residue = set([atom for atom in res.atoms()])
+        bond_partners_in_residue = [item for sublist in [atom.bond_partners for atom in res.atoms()] for item in sublist]
+        if set(atoms_in_residue) != set(bond_partners_in_residue):
+            return False
+    return True
+
+
+def _update_atomtypes(unatomtyped_topology, atomtyped_prototype_topology):
```
justinGilmer: 1 line comment to describe method?

foyer/forcefield.py
```diff
@@ -298,7 +345,26 @@ def createSystem(self, topology, atomtype=True, nonbondedMethod=NoCutoff,
             the newly created System
         """
         if atomtype:
-            find_atomtypes(topology, forcefield=self)
+            independent_residues = _check_independent_residues(topology)
```
summeraz: I like having atomtyping by residue act as the default behavior, but perhaps (as we spoke about regarding unit tests) having a `by_residue` argument to `createSystem()`, or something of that sort would be useful just so that we have the option to revert to the old behavior if desired.  We are already passing `**kwargs` to `createSystem()`, so we should be able to pass `by_residue` straight from `pmd.Structure.apply()` or `mb.Compound.apply()`.

foyer/forcefield.py
```diff
@@ -118,6 +118,53 @@ def _topology_from_parmed(structure, non_element_types):
...
+    topology = app.Topology()
+    chain = topology.addChain()
+    new_res = topology.addResidue(res.name, chain)
+
+    atoms = dict()  # omm.Atom in res : omm.Atom in topology
+
+    for res_atom in res.atoms():
+        topology_atom = topology.addAtom(name=res_atom.name,
+                         #element=elem.Element.getBySymbol(res_atom.name),
+                         element=res_atom.element,
```
summeraz: This line may cause some problems when trying to atomtype coarse-grained systems (where the convention for Foyer is to prepend "element" names with an underscore).  This could be tested with a box of united-atom alkanes since the TraPPE forcefield is already contained in Foyer.  If it turns out there's not a good way around this, we could have `createSystem` check to see if the system contains any coarse-grained particles, in which case the old approach to atomtyping could be used.

foyer/forcefield.py
```diff
@@ -298,7 +345,26 @@ def createSystem(self, topology, atomtype=True, nonbondedMethod=NoCutoff,
...
+                for res in topology.residues():
+                    if res.name not in residue_map.keys():
+                        residue = _topology_from_residue(res)
+                        find_atomtypes(residue, forcefield=self)
+                        residue_map[res.name] = residue
+
+                new_topology = topology
+
+                for key, val in residue_map.items():
+                    new_topology = _update_atomtypes(topology, val)
```
summeraz: I think we could have this just as `_update_atomtypes(topology, val)` and have the `_update_atomtypes` function act directly on the `topology` object, rather than creating additional `topology` instances

summeraz: You could also pass the residue name here as well.  So something like:
```
for res_name, res_template in residue_map.items():
   _update_atomtypes(topology, res_name, res_template)
```
This would eliminate the need to search for the residue name within `_update_atomtypes`.

foyer/forcefield.py
```diff
@@ -298,7 +345,26 @@ def createSystem(self, topology, atomtype=True, nonbondedMethod=NoCutoff,
...
+
+                topology = new_topology
```
summeraz: Again, this should be unnecessary if we operate directly on `topology` inside of `_update_atomtypes`

foyer/tests/test_forcefield.py
```diff
@@ -106,3 +108,13 @@ def test_improper_dihedral():
...
+
+def test_residue_map():
+    ethane = pmd.load_file(get_fn('ethane.mol2'), structure=True)
+    oplsaa = Forcefield(name='oplsaa')
+    topo, NULL = generate_topology(ethane)
+    with_map = pmd.openmm.load_topology(topo,
+        oplsaa.createSystem(topo, use_residue_map = True))
+    without_map = pmd.openmm.load_topology(topo,
+        oplsaa.createSystem(topo, use_residue_map = False))
+    [a.type for a in with_map.atoms] == [a.type for a in without_map.atoms]
```
ctk3b: Could you also add a test where there are bonds between residues?

foyer/forcefield.py
mattwthompson: Inside of ```apply```, ```_topology_from_residue``` acts on a topology that was created with ```generate_topology``` so this case should already be handled.

mattwthompson:
Anything I missed or is this good to go?

chrisiacovella:
Can you do some tests as a function of residue size?  E.g., plot this as a function of total atoms in the system, with different series for say, butane, octane, dodecane, etc. I'm assuming performance will increase as you reduce the overhead related to this operation by considering fewer residues

PR #142, opened 2017-09-18
Add GAFF forcefield

mattwthompson:
I know we're moving toward implementing smaller force fields, but I also think GAFF is a useful one to support.

I grabbed the sans SMARTS file from [here](https://github.com/choderalab/openmoltools/blob/master/openmoltools/parameters/gaff.xml) and set up testing in almost the same way that OPLS was set up. This includes GROMACS topology files from [here](http://virtualchemistry.org/ff.php), specifically the GAFF/GROMACS link.

Felix @ftiet then added SMARTS strings for a handful of ionic liquids and organic molecules. Obviously this only covers a portion of the space that GAFF covers, but it's enough for us to do some science with and anybody can add more strings as needed. Fortunately GAFF only has a few dozen atom types, so it's much less work than OPLS.

Note that partial charges are not included in this XML because AMBER/GAFF assigns charges after atomtyping.


foyer/tests/test_gaff.py
```diff
@@ -0,0 +1,81 @@
...
+                pdb_filename = '{}.pdb'.format(mol_name)
+                top_path = os.path.join(testfiles_dir, mol_name, top_filename)
+                pdb_path = os.path.join(testfiles_dir, mol_name, pdb_filename)
+                structure = pmd.load_file(top_path, xyz=pdb_path, parametrize=False)
+            elif ext == '.mol2':
+                mol2_path = os.path.join(testfiles_dir, mol_name, mol_file)
+                structure = pmd.load_file(mol2_path, structure=True)
+        atomtype(structure, GAFF)
+#
+#    def test_full_parametrization(self):
```
summeraz: Might be a good idea to uncomment and update this test just to make sure the parametrization is working properly, since for the GAFF tests you use a PDB structure file instead of a GRO structure file.

davidlmobley:
Sorry to just occasionally descend out of the ether with very little context, but I'm very interested by this. 

Can I ask how you are able to go from the `gaff.xml` file to something useful? Specifically, how do you encode the atom typing/come up with the SMARTS? This seems to me naively to be sort of a research problem (or a "peer deeply into antechamber where no one dares to look" problem) (AH, update after looking more carefully -- it looks like one of you hand-created SMARTS for just a subset of GAFF. Have you verified that these are correct, e.g. by verifying that one gets the same parameters assigned via your SMARTS as would originally be assigned by, e.g., Antechamber -> tleap?) 

I'd love to see this done someday for all of GAFF/GAFF2, but that's gonna be painful. You mentioned OPLS; have you done this for OPLS?

chrisiacovella:
@davidlmobley At this point we've been doing definitions by hand (for OPLS we've done about 100 parameters from the database in gromacs).  However, we've sort of changed our approach, at least for the time being.  Instead of making a giant library (like converting the complete OPLS library), we've been focusing on smaller forcefield files for specific families (e.g., a forcefield file only for alkanes, one for perfluoroalkanes, etc.), where we've been going back to the original source to get the parameters and their usage (so we can encode the DOI usage in there).  These individual forcefields can be merged into a larger file, but part of the thought processes is that:
-  it's better to have our forcefield files be "compete" (i.e., all SMARTS defined), even those forcefield files cover fewer molecules 
- many of the parameters in those databases aren't very commonly used, so we're sort of focusing on common systems, or at least systems we do
- It makes it easier to check/validate a forcefield file if it doesn't have as many parameters (so the smaller forcefield files makes it easier for validation, especially if doing some of it manually)

We've been having some offline discussions about this and the approach...might be good to have a chat with you in the future about what we've been doing and what is working/not working.

mattwthompson:
Let me take a stab at this, Andrew or Christoph may chime in if I miss or flub something.

> Can I ask how you are able to go from the gaff.xml file to something useful? Specifically, how do you encode the atom typing/come up with the SMARTS? This seems to me naively to be sort of a research problem (or a "peer deeply into antechamber where no one dares to look" problem) (AH, update after looking more carefully -- it looks like one of you hand-created SMARTS for just a subset of GAFF.

Correct, right now we don't have a better way of generating SMARTS than doing it by hand. As you can imagine, this is both a tedious process and one that can be frustrating to do, so we've mostly been adding SMARTS strings on an as-needed basis.

> Have you verified that these are correct, e.g. by verifying that one gets the same parameters assigned via your SMARTS as would originally be assigned by, e.g., Antechamber -> tleap?)

The testing that we do involves reading in a structure file with bond information but no atom types, applying the forcefield (using the XML file of interest) and comparing the atom types that foyer finds to reference topologies that we grabbed from virtualchemistry.org. I don't know how those topology files were generated, perhaps also including an Antechamber test would be useful.

> I'd love to see this done someday for all of GAFF/GAFF2, but that's gonna be painful. You mentioned OPLS; have you done this for OPLS?

Correct, it's no small task. The scope of these SMARTS strings is what is of immediate interest to my research, only a small set of GAFF atom types. We also held off on adding a few atom types we looked at that had confusing behavior.

We have a partially/mostly complete version for OPLS and a somewhat complete version for TraPPE-UA, which you can find [here](https://github.com/mosdef-hub/foyer/tree/master/foyer/forcefields). There are so many obscure and obsolete OPLS-AA atom types that we will probably never hit 100% coverage. Currently the coverage is enough that we are comfortably doing research with it - if you run into any missing atom types we'd be happy to add them in. With OPLS, we have been slowly adding SMARTS strings over time, and I hope to continue that with GAFF.

chrisiacovella:
Here is a link to one of the "small" files (I'll be pushing some changes to this where we define metadata in an xml file, and the readme is automatically generated).  

https://github.com/chrisiacovella/oplsaa_perfluoroalkanes

We'll be creating a lot more of these types of files.

ctk3b:
> Have you verified that these are correct, e.g. by verifying that one gets the same parameters assigned via your SMARTS as would originally be assigned by, e.g., Antechamber -> tleap?)

👍

This PR should definitely add validation against antechamber. That's pretty straight forward and provides a huge safety net.

mattwthompson:
I definitely should have seen that we don't need to use virtualchemistry.org's topologies. Fortunately this means we can draw any molecule we wish and we don't need to be limited by the molecules they selected. Oddly, the big GAFF zip on virtualchemistry.org didn't include simple molecules like benzene, ethane, etc.

Unless we want to make `antechamber` a new dependency, you're going to either have to validate that those files are legit (coordinate file represents the molecule it's named after, sensible atomtypes, etc) or take my word for it (probably not a great idea because I could have easily drawn a molecule incorrectly). 

For each test molecule I changed, I

1. Drew the molecule with Avogadro and saved to pdb.
2. Called `antechamber` to atomtype the molecule:

```
antechamber -fi pdb -fo mol2 -i foo.pdb -o foo.mol2
```

I'm running into some weird issues associated with `parmed` reading chlorine and bromine atoms, as carbon and boron, but I'll push my changes once I figure that out.

ctk3b:
> Unless we want to make antechamber a new dependency, you're going to either have to validate that those files are legit (coordinate file represents the molecule it's named after, sensible atomtypes, etc) or take my word for it (probably not a great idea because I could have easily drawn a molecule incorrectly).

I think it would be just fine to put validation files into this repo and provide the script you used to generate them. 

>Drew the molecule with Avogadro and saved to pdb.
>Called antechamber to atomtype the molecule:
>antechamber -fi pdb -fo mol2 -i foo.pdb -o foo.mol2

You could probably use an already generated library like @davidlmobley's: https://github.com/MobleyLab/FreeSolv

>I'm running into some weird issues associated with parmed reading chlorine and bromine atoms, as carbon and boron, but I'll push my changes once I figure that out.

This is when you load the .mol2 back in after the antechamber step? Not sure what all is contained in those .mol2 files but on parmed's end, this happens when you don't have explicit element information and it's trying to guess an element based on the atom name and e.g. it is not really possible to discern "CL" from "Cl".

davidlmobley:
Let me know if you are interested in FreeSolv and need any info; presumably you could also create subsets which only have specific atom types or only use specific parameters.

mattwthompson:
Here is parmed trying to load in the bromomethane in the FreeSolv database

```
@<TRIPOS>MOLECULE                                                                                                                                                                                                                             
bromomethane
    5     4     1     0     0   
SMALL
No Charge or Current Charge


@<TRIPOS>ATOM
      1 C1          0.1360   -0.1000   -0.0730 c3        1 MOL     -0.039400
      2 Br1        -1.2950    0.9520    0.6960 br        1 MOL     -0.167900
      3 H1          0.8160    0.5790   -0.5900 h1        1 MOL      0.069100
      4 H2          0.6480   -0.6190    0.7390 h1        1 MOL      0.069100
      5 H3         -0.3050   -0.8120   -0.7720 h1        1 MOL      0.069100
@<TRIPOS>BOND
     1    1    2 1   
     2    1    3 1   
     3    1    4 1   
     4    1    5 1   
@<TRIPOS>SUBSTRUCTURE
     1 MOL         1 TEMP              0 ****  ****    0 ROOT
```

```
>>> bmm = pmd.load_file('mol2files_gaff/mobley_8983100.mol2')
>>> [a.name for a in bmm.atoms]
['C1', 'Br1', 'H1', 'H2', 'H3']
>>> [a.element for a in bmm.atoms]
[6, 5, 1, 1, 1]
```

I'll raise an issue with Jason if this should be supported and I'm not doing something silly.

davidlmobley:
Is ParmEd supposed to understand GAFF atom types? I don't normally use its mol2 reader much. I'd naively guess that's the problem you're running into -- that it doesn't know GAFF types (or doesn't assume them unless you ask it to?) so it just takes the first character in the type when it doesn't know and interprets that as the element.

ctk3b:
> that it doesn't know GAFF types (or doesn't assume them unless you ask it to?) so it just takes the first character in the type when it doesn't know and interprets that as the element.

That's precisely the problem. This has come up on a few occasions with the mol2 and gro readers.

davidlmobley:
What are trying to use it for? Can you just use it with the "true" mol2 files in FreeSolv (the ones with SYBYL/TRIPOS types)? Those should parse properly unless you NEED the GAFF types for some reason.

(One thing to keep in mind is that various tools in our field violate the file format specifications of particular formats; putting non-SYBYL atom types in a mol2 file is such a violation. This usually means only specialized tools will read such files. But going back to the correct format -- as I'm proposing in this case -- usually fixes the problem.)

mattwthompson:
We're using it as a correctly-atomtyped template as a reference in tests, so it requires that we read in GAFF atom type information. The tests read in structures, atomtype them with the force field of interest, and compare those generated atom types to whatever reference we provide.

However, this can also be achieved using GROMACS topologies in the FreeSolve library, which don't have the issue of guessing GAFF atom types:

```
>>> bmm = pmd.load_file('gromacs_original/mobley_8983100.top')
>>> [a.element for a in bmm.atoms]
[6, 35, 1, 1, 1]
>>> [a.type for a in bmm.atoms]
['c3', 'br', 'h1', 'h1', 'h1']
>>> 
```

so that's probably the best route to go with this. The test passes using this approach.

mattwthompson:
Okay, that approach is working. I'm going to go back and add a handful of simple molecules to make this more likely to be used. Originally this PR was selfishly focused on molecules I wanted to use (ionic liquids and polar solvents).

For molecules outside of Mobley's set, what should our approach be? I can think of 

1. Go back to drawing by hand
1. Only consider molecules in the Mobley set
1. Push it off to a point in the future

If the Mobley set covers all moieties we'd see in the wild, the second option would probably be best. I'm thinking of ionic liquids, where the molecules are somewhat complicated but the atom types are probably covered by simpler netural molecules. For example, I'd want to do simulations using a 1-ethyl-3-methylimidazolium cation that is obviously not included, but a similar molecule (1-methylimidazole) is, and would probably be sufficient for testing purposes.

@davidlmobley would it be accurate to say your database covers most or all of GAFF? I know that was not why it was constructed, but I would think 500-600 molecules comes close to the 50ish atom types in GAFF.

davidlmobley:
> If the Mobley set covers all moieties we'd see in the wild, the second option would probably be best.

It definitely does not cover all moieties seen in the wild. We have some sets of molecules that include a lot more, plus there are others online. For example, you might look at DrugBank. But, you could also look at our MiniDrugBank, which is basically a subset of DrugBank that has essentially as much chemical diversity. https://github.com/open-forcefield-group/MiniDrugBank

>  would it be accurate to say your database covers most or all of GAFF? I know that was not why it was constructed, but I would think 500-600 molecules comes close to the 50ish atom types in GAFF.

No. There are almost certainly atom types from GAFF which are not used at all, though I don't think we've looked to see which those are. Also, by "covers" do you mean "uses all atom types" or "uses all parameters in..." (the latter is much harder than the former). 

Basically, I can tell you that a FF can cover all of FreeSolv without even coming close to covering all of the chemical space one would want to cover in a "general purpose" small molecule force field. There's a lot of chemistry that just isn't there much at all.

(I have a plot somewhere showing the functional groups represented in DrugBank versus those in FreeSolv and there are various functional groups that just don't show up in FreeSolv at all. LMK if you want to see that.)

mattwthompson:
> Also, by "covers" do you mean "uses all atom types" or "uses all parameters in..." (the latter is much harder than the former).

I mean "use all atom types", I'm sure there are various bond/angle/dihedral terms that would be difficult to hunt down. Even a few of the atom types in GAFF look somewhat esoteric, so my goal is more like "use enough of the atom types to cover a small section of chemical space." I see you cut off molecule size at 100 heavy atoms in MiniDrugBank - here we'd stop around 10.

I think my earlier comment implied I was going for a larger scope than I intend. For comparison, our coverage in OPLS is currently 89 atom types with 73 test molecules - far, far from complete (and as discussed earlier, will probably never be complete) but large enough to do some research with. It's not likely that you'd be able to throw an arbitrary drug molecule at this code and be confident it finds all atom types accurately, but likely that it finds most atom types and any missing or incorrect ones can be corrected for by modifying SMARTS strings or overrides.

My current thinking is that doing a few dozen molecules in the Mobley set would be a sufficient starting point for this PR and be the best way to test the implementation of this force field in foyer - sound good @ctk3b ?

ctk3b:
That sounds like a reasonable start to me but make sure you provide detailed instructions/scripts for how the test data was gathered.

mattwthompson:
I went through and added/modified a few atomtypes and many test molecules. There are still plenty of unsupported GAFF atom types but this is a start. I'm certainly open to doing more if we want to start out with a more complete and more rigorously tested starting point.

For each test molecule, I looked to see if it was in the `database.txt` included in [FreeSolv](https://github.com/MobleyLab/FreeSolv/blob/master/database.txt). These lines look like this

```
mobley_2068538; CCC; propane; 2.00; 0.20; 2.50; 0.02; 10.1039/P29900000291; 10.1021/acs.jced.7b00104; Experimental uncertainty as suggested by 10.1039/P29900000291 -- 0.2 kcal/mol.  
```

where only the first ("mobley_name") and third (common name) columns are important for our purposes. I then copied the `.top` and `.gro` files into the testing directory:

```
mkdir propane
cp ~/software/FreeSolv/gromacs_original/mobley_2068538.* propane/
```

and added `propane` to `implemented_gaff_tests.txt`.

I did not change the file names to `propane.top` and `propane.gro`. I also did not automate this process; perhaps that would be useful if we were looking to add all or most of the Mobley set, but I'm not set on doing that. There are probably plenty functional groups and drug molecules that would cause us headaches.

davidlmobley:
Incidentally, I recommend never using compound names as filenames as they can contain italics, french braces, commas, spaces, and other commas which are a real pain to deal with.

davidlmobley:
(directory names either, for that matter).

PR #148, opened 2017-10-20
NetworkX migration

summeraz:
Several functions and function names have changed from NetworkX 1.0 to NetworkX 2.0. For example, the `Graph.nodes_iter()` function has been replaced by `Graph.nodes()`. This PR changes our NetworkX syntax to conform to both versions.

### DIFF
--- a/foyer/smarts_graph.py
+++ b/foyer/smarts_graph.py
@@ -154,7 +154,7 @@ class SMARTSGraph(nx.Graph):
         When this function gets used in atomtyper.py, we actively modify the
         white- and blacklists of the atoms in `topology` after finding a match.
         This means that between every successive call of
-        `subgraph_isomorphisms()`, the topology against which we are
+        `subgraph_isomorphisms_iter()`, the topology against which we are
         matching may have actually changed. Currently, we take advantage of this
         behavior in some edges cases (e.g. see `test_hexa_coordinated` in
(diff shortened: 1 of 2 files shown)


foyer/smarts_graph.py
```diff
@@ -191,11 +191,10 @@ def find_matches(self, topology):
 
         # The first node in the smarts graph always corresponds to the atom
         # that we are trying to match.
-        first_atom = next(self.nodes_iter())
```
summeraz: @ctk3b Does this change make sense (along with the change to line 197/198)? This doesn't seem quite right, but all of our tests are passing...

ctk3b: That might just be a vestige from an earlier time when we were still trying to use the actual `atom` objects as the node items. [These are now integers and the atom is the node attribute.](https://github.com/mosdef-hub/foyer/blob/master/foyer/smarts_graph.py#L53)

Easy check: set a breakpoint at 193 and see what nodes looks like. If it's just an integer then what you did is correct and I would move the comment in 192 down to 197.

summeraz: Yes, I checked this before I made the change and it seemed that `first_atom` was always an integer, specifically `0`. Sounds likely this change is okay then. I'll move that comment down.

PR #151, opened 2017-12-14
Allow single-atom residues to use the residue map

mattwthompson:
The `_check_independent_residues` function, which needs to be `True` in order to use the residue map feature I added a while back, breaks if the atoms in the residue have no bond partners. The function checks to see if any bond partners of atoms in a given residue are not themselves atom in said residue:

```
if set(atoms_in_residue) != set(bond_partners_in_residue):
    return False
```
But this was breaking when `bond_partners_in_residue` is an empty list.

This small fix handles this case, and should allow single-atom residues to use the residue map for speedup in atomtyping.

h/t @justinGilmer, who found this by way of an argon system that was slow to apply a force field to.

### DIFF
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -162,7 +162,7 @@ def _check_independent_residues(topology):
         atoms_in_residue = set([atom for atom in res.atoms()])
         bond_partners_in_residue = [item for sublist in [atom.bond_partners for atom in res.atoms()] for item in sublist]
         # Handle the case of a 'residue' with no neighbors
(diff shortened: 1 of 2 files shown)


ctk3b:
Nice LGTM!

Looks like Travis is having a blast again though... can you try removing the `rvm` update? I think we might have only needed the `brew update`since that worked in a recent mdtraj PR but @justinGilmer probably knows more.

mattwthompson:
👍 

Will merge #152 into this once those tests pass

PR #152, opened 2017-12-14
Attempt to fix `shell session update` error

justinGilmer:
It appears that travis is having some
issues again.

This is an issue with `rvm` hopefully this fixes it.


justinGilmer:
Still a mystery to me, but it works.  @mattwthompson Once this is merged, feel free to pull this into your PR and we can see if the tests pass.

PR #155, opened 2018-01-09
Checks for correct topology parameterization

summeraz:
Currently if parameters are not found for certain topology objects (angles, proper dihedrals, and improper dihedrals) Foyer will continue through without notifying the user. This can lead to serious problems as the incorrect parameterization may not be caught until after production runs have been performed on systems parameterized by Foyer. In many cases, parameters are not found due to typos in the user's force field XML file, although there are cases (e.g. coarse-grained systems) where dihedral forces may not be desired for atoms separated by three bonds. I've implemented a fix here that provides either a warning or error to the user if the systems is found to be incorrectly parameterized. By default, an error is raised if angles and proper dihedrals are not parameterized correctly and a warning is raised if impropers are not parameterized correctly. This behavior can be overridden by the user.


foyer/forcefield.py
```diff
@@ -317,9 +328,42 @@ def apply(self, topology, references_file=None, *args, **kwargs):
...
+            if assert_dihedral_params:
+                raise Exception(msg)
+            else:
+                warnings.warn(msg)
+        if data.impropers and (len(data.impropers) != len(structure.impropers)):
+            msg = ("Parameters have not been assigned to all impropers. Total "
+                   "system impropers: {}, Parameterized impropers: {}"
+                   "".format(len(data.impropers), len(structure.impropers)))
+            if assert_improper_params:
+                raise Exception(msg)
```
ctk3b: Could you factor these 3 checks out into a consistent "error or warn" function? Lots of repeated code here

foyer/forcefield.py
```diff
@@ -589,7 +633,7 @@ def createSystem(self, topology, atomtype=True, use_residue_map=True,
         # Execute scripts found in the XML files.
         for script in self._scripts:
             exec(script, locals())
-        return sys
+        return sys, data
```
ctk3b: nit: Is there any other way to reach this data? Perhaps setting it as an attribute? This function is called `createSystem` so it's somewhat unintuitive that it would return other things as well.

foyer/tests/test_forcefield.py
```diff
@@ -144,3 +144,19 @@ def test_independent_residues_atoms():
...
+        ethane = oplsaa_angle_typo.apply(ethane)
+    with pytest.warns(UserWarning):
+        ethane = oplsaa_angle_typo.apply(ethane, assert_angle_params=False)
+
+    ethane = mb.load(get_fn('ethane.mol2'))
+    oplsaa_dihedral_typo = Forcefield(forcefield_files=get_fn('ethane-dihedral-typo.xml'))
+    with pytest.raises(Exception):
+        ethane = oplsaa_dihedral_typo.apply(ethane)
+    with pytest.warns(UserWarning):
+        ethane = oplsaa_dihedral_typo.apply(ethane, assert_dihedral_params=False)
```
ctk3b: 👍 

Could you factor out this test as well and just `@parameterize` the function with the forcefield file name and the assert flag?

foyer/forcefield.py
summeraz: Good call, should help to condense this

foyer/forcefield.py
summeraz: Something did feel a little off when I took this approach... `data` is only available from within `createSystem`, but yeah I could just throw a line like `self._system_data = data` or something like that before the return.

foyer/tests/test_forcefield.py
summeraz: Sure, that should make the tests a bit cleaner

summeraz:
Will work on getting those changes made. And yes, unparameterized/missing bonds already yield an error, although the message itself is not particularly helpful (something about an empty array or similar, so not mentioning the missing bonds explicitly), but occurs in OpenMM-land and would be difficult for us to catch since that's where the parameterization happens.

summeraz:
Also, it looks like there are several molecules in the TraPPE and OPLS tests where parameters must not be defined for all relevant angles or dihedrals as some of these tests are failing after hitting the checks I've put in. So this is something we'll need to look into...

mattwthompson:
This is exactly the problem that I encountered when using foyer with an xml I built from the ground up: some dihedrals weren't getting assigned (in one case, a very important one) but nothing in my workflow was flagging it and it went unnoticed for several months and many GFlops of simulations

foyer/forcefield.py
```diff
@@ -337,32 +353,24 @@ def apply(self, topology, references_file=None, assert_angle_params=True,
         have parameters assigned. OpenMM will generate an error if bond parameters
         are not assigned.
         '''
+        data = self._SystemData
```
ctk3b: 👍

foyer/forcefield.py
```diff
@@ -436,26 +444,25 @@ def createSystem(self, topology, atomtype=True, use_residue_map=True,
             else:
                 find_atomtypes(topology, forcefield=self)
 
-        data = app.ForceField._SystemData()
-        data.atoms = list(topology.atoms())
-        for atom in data.atoms:
-            data.excludeAtomWith.append([])
+        self._SystemData.atoms = list(topology.atoms())
```
ctk3b: Could you limit this to just adding a single line here which sets `data = self._SystemData()` in place of the existing `data = app.ForceField._SystemData()`?

summeraz:
I need to look into the ParmEd source a bit deeper, but it looks like even when improper dihedrals are present `Structure.impropers` is not always populated. For now I am counting the total number of impropers by taking the sum of the lengths of `Structure.impropers` and `[dihedral for dihedral in structure.dihedrals if dihedral.improper]`. However, I'm not sure if a scenario could exist where this would result in double counting.

summeraz:
Discovered a problem with saving topological data from `createSystem` to the newly added `_SystemData` object. If a forcefield is used to atomtype multiple systems the number of topology object (e.g. bonds, angles, etc.) is additive rather than unique to each system. [My latest commit](https://github.com/summeraz/foyer/blob/6425334f520679a6f62389833a4427eea4663c05/foyer/forcefield.py#L457) solves this, but seems pretty hacky, so I'm guessing we can find a better solution.

summeraz:
It turns out that several of atomtypes we have SMARTS strings defined for in the OPLS force field XML did not have the `class` attribute assigned properly (specifically atomtypes for alkynes, pyridine, pyrimidine, and acetals). I've added the correct classes and this has reduced the number of molecule tests where the number of parametrized angles/dihedrals does not match expectations. However, there are still 12 molecules that have atomtypes assigned properly, but do not have parameters assigned for all angles/dihedrals. Specifically:

- **1-cyclopropylpropane** (*missing an angle parameter*)
- **1-nitropropane** (*missing a dihedral parameter*)
- **2-propylpyrrole** (*missing a dihedral parameter*)
- **24-pentanedione** (*missing an angle parameter*)
- **E-hex-2-ene** (*missing a dihedral parameter*)
- **dimethyl-sulfoxide** (*missing all dihedral parameters*)
- **ethylene-carbonate** (*missing an angle parameter*)
- **formaldehyde** (*missing an angle parameter*)
- **paraldehyde** (*missing angle parameters*)
- **pentanenitrile** (*missing a dihedral parameter*)
- **propylene-carbonate** (*missing an angle parameter*)
- **styrene** (*missing dihedral parameters*)

Many of these molecules are functionalized alkanes, where parameters are missing for an angle or dihedral that involves the α, β, and γ carbons. In these cases the parameters simply do not exist in the OPLS forcefield shipped with GROMACS (which we converted to this forcefield XML). Many of these missing dihedrals are C-CT-CT-CT dihedrals where the CT-CT-CT-CT dihedral parameters would probably be sufficient (as OPLS uses in many other cases), but we would leave it up to the user to make this decision and add these parameters themselves, keeping the XML shipped with Foyer the same as what is contained in GROMACS. I'm not sure there's really much else we can do. @justinGilmer and myself are working on a website for Foyer that will contain more comprehensive documentation than is available on Github, and we plan to fully document these limitations.

summeraz:
I've removed secondary alcohols from our list of molecules successfully parameterized using our TraPPE forcefield XML, as these feature a dihedral functional form that we do not currently support.

foyer/tests/test_forcefield.py
```diff
@@ -144,3 +144,16 @@ def test_independent_residues_atoms():
...
+    ("ethane-dihedral-typo.xml", {"assert_dihedral_params": False})
+])
+def test_missing_topo_params(ff_filename, kwargs):
+    """Test that the user is notified if not all topology parameters are found."""
+    ethane = mb.load(get_fn('ethane.mol2'))
+    oplsaa_with_typo = Forcefield(forcefield_files=get_fn(ff_filename))
+    with pytest.raises(Exception):
+        ethane = oplsaa_with_typo.apply(ethane)
+    with pytest.warns(UserWarning):
+        ethane = oplsaa_with_typo.apply(ethane, **kwargs)
```
ctk3b: :+1:

ctk3b:
> I need to look into the ParmEd source a bit deeper, but it looks like even when improper dihedrals are present Structure.impropers is not always populated. For now I am counting the total number of impropers by taking the sum of the lengths of Structure.impropers and [dihedral for dihedral in structure.dihedrals if dihedral.improper]. However, I'm not sure if a scenario could exist where this would result in double counting.

This sounds like it's worth understanding properly. Is it a specific type of improper that causes the problem? I can help debug if needed

summeraz:
Here's [a ZIP file](https://github.com/mosdef-hub/foyer/files/1791231/improper-test.zip) with a minimal example that reproduces the strange storage behavior for impropers which includes a dummy forcefield file, a PDB file for an ethene molecule (which should have two impropers), and a script that takes the PDB and atomtypes it. Here's the atomtyping script:
```
from foyer import Forcefield
import parmed as pmd 

ff = Forcefield(forcefield_files='dummy-ff.xml')
structure = pmd.load_file('ethene.mol2', structure=True)
atomtyped_structure = ff.apply(structure)
```
And here's output from inspection of `atomtyped_structure`:
```
(Pdb) len(atomtyped_structure.impropers)
0
(Pdb) len([dihedral for dihedral in atomtyped_structure.dihedrals if dihedral.improper])
2
```
So the two impropers are correctly assigned, but they are stored within `Structure.dihedrals` (where the `improper` attribute is set to `True`) rather than being stored within `Structure.impropers`. I'll look through the ParmEd source to try to figure out why this happens. Perhaps `Structure.impropers` is some sort of legacy attribute that is no longer used.

summeraz:
Okay so here is what I've found from my digging so far:

The relevant function in the ParmEd source is [load_topology](https://github.com/ParmEd/ParmEd/blob/74a84d30eadd2a0a2bf1e17f6612b3409f341a3e/parmed/openmm/topsystem.py#L30), where an OpenMM `System` and `Topology` are used to create a ParmEd `Structure`. The relevant OpenMM Forces appear to be [PeriodicTorsionForce](https://github.com/ParmEd/ParmEd/blob/74a84d30eadd2a0a2bf1e17f6612b3409f341a3e/parmed/openmm/topsystem.py#L167-L168), [RBTorsionForce](https://github.com/ParmEd/ParmEd/blob/74a84d30eadd2a0a2bf1e17f6612b3409f341a3e/parmed/openmm/topsystem.py#L169-L170), and [CustomTorsionForce](https://github.com/ParmEd/ParmEd/blob/74a84d30eadd2a0a2bf1e17f6612b3409f341a3e/parmed/openmm/topsystem.py#L171-L174). [Per the OpenMM documentation](http://docs.openmm.org/6.2.0/userguide/application.html), child definitions for each of these forces can be of either "proper" or "improper" type. However, in ParmEd's `load_topology` each of these three torsion forces is handled differently:

- For `PeriodicTorsionForce` [improper torsions are added to `Structure.dihedrals` with `improper=True`](https://github.com/ParmEd/ParmEd/blob/74a84d30eadd2a0a2bf1e17f6612b3409f341a3e/parmed/openmm/topsystem.py#L269-L272)
- For `RBTorsionForce`, [there appears to be no distinction between improper and proper torsions, as these are all added to `Structure.rb_torsions`](https://github.com/ParmEd/ParmEd/blob/74a84d30eadd2a0a2bf1e17f6612b3409f341a3e/parmed/openmm/topsystem.py#L295)
- For `CustomTorsionForce`, [improper torsions are added to `Structure.impropers` *if* they are of a specific functional form](https://github.com/ParmEd/ParmEd/blob/74a84d30eadd2a0a2bf1e17f6612b3409f341a3e/parmed/openmm/topsystem.py#L336)

What I gather from this is that there is no way from just the data in a ParmEd `Structure` to know the number of proper and improper torsions in a system because if using a Ryckaert-Bellemans functional form these are all lumped together into a single attribute (`Structure.rb_torsions`). However, I still think it is good for us to have some sort of check on our end, so I think the best solution is to keep the approach that I've currently taken with the caveat that counts will be incorrect if improper torsions are defined within the `RBTorsionForce` section of your XML file. As long as this is well-documented, I think this is a better solution than not having any sanity checks at all as the user still has the ability to override these assertions.

ctk3b:
Does parmed hash dihedral types based on their contents? That would be an easy way to avoid double counting.

More generally, I can't think of a scenario where a forcefield would define the exact same dihedral in two different ways. I would expect that to be an incorrectly constructed FF file. Ideally we would of course not allow this to silently slip by so a double counting check would be helpful but perhaps not absolutely necessary.

summeraz:
Right, I no longer think double counting will be a problem. Before looking into the source I was thinking perhaps it was possible for an improper dihedral to exist within both `Structure.impropers` and `Structure.dihedrals`, but this does not appear to be possible. So the only remaining issue is that it is ambiguous concerning whether each of the items in `Structure.rb_torsions` is a proper or improper dihedral. To which I think the best solution is to assume all of these are proper dihedrals (making sure we have this well-documented). If a user does have improper dihedrals within the RBTorsionForce portion of their XML, the sanity check for the number of parameterized proper dihedrals will cause `apply` to exit with an error. The user can then manually override this behavior by passing `assert_dihedral_params=False` to `apply`.

ctk3b:
>  I think the best solution is to assume all of these are proper dihedrals 

👍 

While it's perfectly possible, I don't think I've seen a forcefield implement impropers using this functional form.

PR #158, opened 2018-03-05
Add DOIs for OPLS atomtypes

summeraz:
Our OPLS force field XML (which we created using atomtype definitions and parameters from GROMACS) features several atomtypes with defined SMARTS strings that lack a DOI reference. This PR adds DOIs to these atomtypes so that the source of the relevant parameters is more transparent.


summeraz:
I'm not finished with this yet, but wanted to go ahead and open the PR so that I can log comments as to issues I run into trying to locate the proper DOI references.

summeraz:
The [sulfoxide OPLS parameters](https://github.com/summeraz/foyer/blob/50d3acfccb25099ad2eaf515dbf9e5b7cea777eb/foyer/forcefields/oplsaa.xml#L496-L498) do not appear to have a published source. The GROMACS parameters that we have in our XML appear in a [recent OPLS paper](https://doi.org/10.1021/acs.jpcb.7b04233), where these are compared with parameters derived for the improved OPLS-AAx force field. It seems most papers in the literature that used these parameters link to [this paper](https://doi.org/10.1021/ja9539195), where the parameters are listed in Ref. 18, although the charges used here are different than those in our XML and as listed in [the 2017 paper](https://doi.org/10.1021/acs.jpcb.7b04233). I went ahead and added both DOIs to this atomtype. I do not believe our XML parser is currently designed to take more than a single DOI, so I will have to update that.

Also, it might be useful to add a `note` attribute to `Type`s within the `AtomTypes` section (unless we already have the ability to do this), so that messages (like the above) could be added if citations are not straightforward.

summeraz:
Parameters for the [α-carbon in alkyl pyrroles](https://github.com/summeraz/foyer/blob/50d3acfccb25099ad2eaf515dbf9e5b7cea777eb/foyer/forcefields/oplsaa.xml#L664-L665) do not appear to have a published source. It appears with OPLS that for most α-carbons the epsilon and sigma used are the same as for alkane carbons and the charge is simply adjusted to preserve molecular charge neutrality. As per my above comment, this is something we could probably just add to a `note` attribute and link the DOI to the alkane carbon parameters.

summeraz:
Parameters for [cyclopropanes](https://github.com/summeraz/foyer/blob/50d3acfccb25099ad2eaf515dbf9e5b7cea777eb/foyer/forcefields/oplsaa.xml#L682-L684) do not appear to have a published source. However, these parameters are identical to those for alkanes. Bonded parameters are obviously different; however, for now I'm just focusing on nonbonded parameters since we do not yet have the ability to add DOI tags to bonded parameters.

summeraz:
DOIs are still needed for [propylene and ethylene carbonate](https://github.com/summeraz/foyer/blob/50d3acfccb25099ad2eaf515dbf9e5b7cea777eb/foyer/forcefields/oplsaa.xml#L739-L747) parameters and for [alkyne](https://github.com/summeraz/foyer/blob/50d3acfccb25099ad2eaf515dbf9e5b7cea777eb/foyer/forcefields/oplsaa.xml#L780-L785) parameters.

mattwthompson:
And to think there was a time I thought adding SMARTS strings was the hard part!

Issue #157, opened 2018-02-28
Forcefield.createSystem cannot accept atomtype=False


mattwthompson:
I was messing around with `forcefield.createSystem` to see how much time is taken up by all of the steps involved `openmm.System` compared to the atomtyping step(s). I wanted to use `atomtype=False` to skip over the atomtyping but found this is not a valid argument. The [block](https://github.com/mosdef-hub/foyer/blob/master/foyer/forcefield.py?utf8=%E2%9C%93#L373) at L373 gets skipped over and the function eventually errors out at L428 when it fails to find atomtypes. This is not tested for, and from what I can tell will fail for any system lacking nonbonded parameters.

Maybe this should only be called on a `openmm.Topology` that already has the appropriate nonbonded information, but I don't see how that can happen - nowhere else in `forcefield.py` is `find_atomtypes` called. Perhaps this was saved for a future feature? I'm confused how this case could even happen if `createSystem` returns `openmm.System` and there is currently no other operation on `openmm.Topology` that would fill in that information.

We should either document this or remove `atomtype` as an argument altogether.

mattwthompson:
Another suggestion - what if we wrap the `if Atomtype` block into a separate function?

ctk3b:
> This is not tested for, and from what I can tell will fail for any system lacking nonbonded parameters.

Good catch. Should definitely have been tested but as you point out is a flawed concept to begin with. IIRC, in the standard openmm createSystem, this is the place where templates are applied -something we would [like to support](https://github.com/mosdef-hub/foyer/issues/80) as well. Presumably, there should be some check post atomtyping/templating to assert that all atoms have a type. If they don't, there's not enough information to build a system.

> Another suggestion - what if we wrap the if Atomtype block into a separate function?

That seems sensible regardless of the above issues. +1 from me.

mattwthompson:
👍 I'll see if I can wrangle a PR to at least address the second point

mattwthompson:
Resolved #159 

* Atomtyping moved into a new function

* `atomtype` removed as an argument to `Forcefield.createSystem`

PR #159, opened 2018-03-06
Move atomtyping step into a specific function

mattwthompson:
Dicussed in #157, but to repeat the important bit:

Currently there's a bit of code at the beginning of `Forcefield.createSystem` that manages the atomtyping step, whereas this should more cleanly done by being sectioned off to a different function.

* I'm open to naming the function something more elegant, I'm not happy with `Forcefield.run_atomtyping` but couldn't come up with something better.

* I'm not sure if it should be a helper function, or whatever `Forcefield._run_atomtyping` would be called. It seems possible for a user to want to call it directly without creating the entire `openmm.System` or `parmed.Structure`.

* I don't think a test is necessary for this, as I'm just moving code around. But I also couldn't think of how a useful test would be structured. I could verify that the returned `openmm.Topology` has atomtype information, but this should cause later tests to crash anyway.

* I added `use_residue_map` as an argument to `Forcefield.apply`. I don't know why it wasn't there before, and don't see any reason for it to be buried inside these other functions.

### DIFF
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -309,7 +309,14 @@ class Forcefield(app.ForceField):
             Name of file where force field references will be written (in Bibtex
             format)
         use_residue_map : boolean
-            Bypass atomtyping duplicate residues using a residue map
+            Store atomtyped topologies of residues to a dictionary that maps
+            them to residue names.  Each topology, including atomtypes, will be
+            copied to other residues with the same name. This avoids repeatedly
+            calling the subgraph isomorphism on idential residues and should
+            result in better performance for systems with many identital
+            residues, i.e. a box of water. Note that for this to be applied to
+            independent molecules, they must each be saved as different
+            residues in the topology.
         """
         if not isinstance(topology, app.Topology):
             residues = kwargs.get('residues')
@@ -341,7 +348,14 @@ class Forcefield(app.ForceField):
         topology : openmm.app.Topology
             Molecular structure to find atom types of
         use_residue_map : boolean
-            Bypass atomtyping duplicate residues using a residue map
+            Store atomtyped topologies of residues to a dictionary that maps
+            them to residue names.  Each topology, including atomtypes, will be
+            copied to other residues with the same name. This avoids repeatedly
+            calling the subgraph isomorphism on idential residues and should
+            result in better performance for systems with many identital
+            residues, i.e. a box of water. Note that for this to be applied to
+            independent molecules, they must each be saved as different
+            residues in the topology.
         """
         if use_residue_map:
             independent_residues = _check_independent_residues(topology)
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -383,8 +383,7 @@ class Forcefield(app.ForceField):
 
         return topology
 
-    def createSystem(self, topology, atomtype=True, use_residue_map=True,
-                     nonbondedMethod=NoCutoff,
+    def createSystem(self, topology, nonbondedMethod=NoCutoff,
                      nonbondedCutoff=1.0 * u.nanometer, constraints=None,
                      rigidWater=True, removeCMMotion=True, hydrogenMass=None,
                      **args):
@@ -394,8 +393,6 @@ class Forcefield(app.ForceField):
         ----------
         topology : Topology
             The Topology for which to create a System
-        use_residue_map : boolean
-            Bypass atomtyping duplicate residues using a residue map
         nonbondedMethod : object=NoCutoff
             The method to use for nonbonded interactions.  Allowed values are
             NoCutoff, CutoffNonPeriodic, CutoffPeriodic, Ewald, or PME.
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -330,6 +330,31 @@ class Forcefield(app.ForceField):
 
         return structure
 
+    def run_atomtyping(self, topology, use_residue_map):
+
+        if use_residue_map:
+            independent_residues = _check_independent_residues(topology)
+
+            if independent_residues:
+                residue_map = dict()
+
+                for res in topology.residues():
+                    if res.name not in residue_map.keys():
+                        residue = _topology_from_residue(res)
+                        find_atomtypes(residue, forcefield=self)
+                        residue_map[res.name] = residue
+
+                for key, val in residue_map.items():
+                    _update_atomtypes(topology, key, val)
+
+            else:
+                find_atomtypes(topology, forcefield=self)
+
+        else:
+            find_atomtypes(topology, forcefield=self)
+
+        return topology
+
     def createSystem(self, topology, atomtype=True, use_residue_map=True,
                      nonbondedMethod=NoCutoff,
                      nonbondedCutoff=1.0 * u.nanometer, constraints=None,
@@ -370,27 +395,6 @@ class Forcefield(app.ForceField):
         system
             the newly created System
         """
-        if atomtype:
-            if use_residue_map:
-                independent_residues = _check_independent_residues(topology)
-
-                if independent_residues:
-                    residue_map = dict()
-
-                    for res in topology.residues():
-                        if res.name not in residue_map.keys():
-                            residue = _topology_from_residue(res)
-                            find_atomtypes(residue, forcefield=self)
-                            residue_map[res.name] = residue
-
-                    for key, val in residue_map.items():
-                        _update_atomtypes(topology, key, val)
-
-                else:
-                    find_atomtypes(topology, forcefield=self)
-
-            else:
-                find_atomtypes(topology, forcefield=self)
 
         data = app.ForceField._SystemData()
         data.atoms = list(topology.atoms())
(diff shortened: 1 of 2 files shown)


foyer/forcefield.py
```diff
@@ -308,6 +308,8 @@ def apply(self, topology, references_file=None, *args, **kwargs):
         references_file : str, optional, default=None
             Name of file where force field references will be written (in Bibtex
             format)
+        use_residue_map : boolean
+            Bypass atomtyping duplicate residues using a residue map
```
ctk3b: Could you maybe clarify a bit more what this does via the variable name and/or the description here? Maybe something along the lines of `copy_atomtypes_to_duplicate_residues`

mattwthompson:
Do you want this in a test or also in the function itself?

mattwthompson:
I don't think the docstring I added is great (please suggest fixes/improvements!) but it's more descriptive. Personally I like the name of the argument - it's not great that it describes a variable that's totally internal -
 but it's a difficult feature to describe in an argument alone.

mattwthompson:
And coveralls is whining that I added one un-tested line but only three tested lines 🙄

ctk3b:
Doc string LGTM! Would be easy to add a test case that triggers the error to get that coverage up 😁

foyer/forcefield.py
```diff
@@ -308,6 +308,15 @@ def apply(self, topology, references_file=None, *args, **kwargs):
         references_file : str, optional, default=None
             Name of file where force field references will be written (in Bibtex
             format)
+        use_residue_map : boolean
```
summeraz: A couple nitpicks in this docstring. Could you make this `use_residue_map : boolean, optional, default=True`?

foyer/forcefield.py
```diff
@@ -308,6 +308,15 @@ def apply(self, topology, references_file=None, *args, **kwargs):
...
+            Store atomtyped topologies of residues to a dictionary that maps
+            them to residue names.  Each topology, including atomtypes, will be
+            copied to other residues with the same name. This avoids repeatedly
+            calling the subgraph isomorphism on idential residues and should
```
summeraz: Typo (should be `identical`)

foyer/forcefield.py
```diff
@@ -308,6 +308,15 @@ def apply(self, topology, references_file=None, *args, **kwargs):
...
+            result in better performance for systems with many identital
```
summeraz: Typo (should be `identical`)

foyer/forcefield.py
```diff
@@ -317,6 +326,7 @@ def apply(self, topology, references_file=None, *args, **kwargs):
             positions = np.empty(shape=(topology.getNumAtoms(), 3))
             positions[:] = np.nan
         box_vectors = topology.getPeriodicBoxVectors()
+        topology = self.run_atomtyping(topology, use_residue_map=use_residue_map)
```
summeraz: 👍

foyer/forcefield.py
```diff
@@ -330,6 +340,49 @@ def apply(self, topology, references_file=None, *args, **kwargs):
...
         return structure
 
+    def run_atomtyping(self, topology, use_residue_map=True):
+        """Atomtype the topology
+
+        Parameters
+        ----------
+        topology : openmm.app.Topology
+            Molecular structure to find atom types of
+        use_residue_map : boolean
```
summeraz: Ditto the comments on the above docstring

foyer/forcefield.py
```diff
@@ -330,6 +340,49 @@ def apply(self, topology, references_file=None, *args, **kwargs):
...
+
+        else:
+            find_atomtypes(topology, forcefield=self)
+
+        if not all([a.id for a in topology.atoms()][0]):
+            raise ValueError('Not all atoms in topology have atom types')
+
+        return topology
+
     def createSystem(self, topology, atomtype=True, use_residue_map=True,
```
summeraz: We should be able to remove the `atomtype` and `use_residue_map` arguments now. In fact, is `createSystem` now exactly the same as OpenMM's? If so, we could actually delete this entire function... (although this may cause problems with #155). Thoughts @ctk3b?

ctk3b: Double check me but I think the base method includes the logic for template based atomtyping

foyer/forcefield.py
mattwthompson: I should probably have a vim plugin that prevents me from making these sorts of mistakes....

foyer/forcefield.py
mattwthompson: I'll clean up the arguments but this and a few other ideas with `Forcefield.createSystem` vs OpenMM's original function should be dealt with at some point. There's an open issue about templating or something that I can't find at the moment but should be referenced.

mattwthompson:
I'm having trouble seeing a good way to test for missing atomtypes. If an atom in the topology truly lacks atomtypes defined in the forcefield, I think errors in `find_atomtypes` would be raised. I could try to manually pop `a.id` from a random atom to force it, I guess?

foyer/forcefield.py
mattwthompson: We think you're right in saying that they do their reside teamplate-based atomtyping in their `Forcefield.createSystem`. It doesn't appear we can use their base method out of the box.

Worth noting that their while templates are much more involved, we are already doing some work in creating a map between residue/molecule names to atomtyped topologies but trashing it once the system is atomtyped. I don't know if it's feasible or worth considering trying to save `residue_map` to disk? Or maybe converting it into a single large `residueTemplate` that could then be passed to OpenMM's base method?

Here's their method at a point in history that should be close to when we grabbed it:

https://github.com/pandegroup/openmm/blob/c9fcabb5da87ef61ac0c777423b1f3e1d96e654e/wrappers/python/simtk/openmm/app/forcefield.py#L1041

Also notable that there have been some updates to that block of in the past year. Probably not much of consequence now but something to worry about later, especially if we want great integration with OpenMM ...

ctk3b:
https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

mattwthompson:
In this case I just don't know how to raise that error. I get how to use `pytest` to verify that a desired error has been raised.

ctk3b:
> don't know how to raise that error 

I would just call createSystem without the atomtyping flag and it should yell at you

mattwthompson:
But that argument is gone now (the atomtyping happens before `createSystem` is called).

We can try applying an inappropriate force field to a molecule, 
but that triggers errors elsewhere that are already test for:

```
    def test_missing_atomtypes():
        ethane = pmd.load_file(get_fn('pf6.mol2'), structure=True)
        oplsaa = Forcefield(name='oplsaa')
        topo, NULL = generate_topology(ethane)
        with pytest.raises(ValueError):
>           oplsaa.run_atomtyping(topo)

foyer/tests/test_forcefield.py:137: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
foyer/forcefield.py:369: in run_atomtyping
    find_atomtypes(residue, forcefield=self)
foyer/atomtyper.py:43: in find_atomtypes
    _resolve_atomtypes(topology)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

topology = <Topology; 1 chains, 1 residues, 7 atoms, 6 bonds>

    def _resolve_atomtypes(topology):
        """Determine the final atomtypes from the white- and blacklists. """
        for atom in topology.atoms():
            atomtype = [rule_name for rule_name in atom.whitelist - atom.blacklist]
            if len(atomtype) == 1:
                atom.id = atomtype[0]
            elif len(atomtype) > 1:
                raise FoyerError("Found multiple types for atom {} ({}): {}.".format(
                    atom.index, atom.element.name, atomtype))
            else:
                raise FoyerError("Found no types for atom {} ({}).".format(
>                   atom.index, atom.element.name))
E               foyer.exceptions.FoyerError: Found no types for atom 0 (phosphorus).

foyer/atomtyper.py:104: FoyerError

```

ctk3b:
Ok good point - sorry I misread how you had implemented this. I think we're good then since we already guarantee that all atoms got a type before calling `createSystem` which is the whole point

mattwthompson:
One of the travis builds on mac got hung up on something, but the second most recent test did not, and the most recent commit is minor. Should be fine.

Somebody else should give it a once-over before merging, but from my perspective this is good to go.

ctk3b:
Do you have admin access on travis? To restart builds etc?

mattwthompson:
No on upstream I don't think

mattwthompson:
Did I miss anything or is this good to go?

summeraz:
LGTM. @ctk3b anything else, or are we good to merge?

ctk3b:
Sorry I missed this - LGTM!

PR #160, opened 2018-03-06
Assign parameter precedence by number of `type` attributes

### LINKED ISSUE #135: Precedence of torsions defined using "type" vs. "class"
Within a force field XML file, torsions (and I'm pretty sure bonds and angles as well) can be defined using either "class" notation, e.g.
`<Proper class1="CT" class2="CT" class3="CT" class4="CT" c0="2.9288" c1="-1.4644" c2="0.2092" c3="-1.6736" c4="0.0" c5="0.0"/>`
or "type" notation, e.g.
`<Proper type1="opls_961" type2="opls_136" type3="opls_136" type4="opls_136" c0="-0.987424" c1="0.08363" c2="-0.08368" c3="-0.401664" c4="1.389088" c5="0.0"/>`

The practical use of having both types of torsions defined would be to have the more general case be covered using the "class" notation and defining more specific dihedrals using the "type" notation.  It seems to me then that dihedrals using "type" notation should essentially override those defined using the "class" notation; however, the current functionality simply assigns to a dihedral whichever parameters are listed first in the XML. The dihedral parameters are actually assigned using OpenMM, so I'm not sure if this precedence is something they would want or not, but is something I think we should at least implement in Foyer.

summeraz:
As described in #135, parameters for topology objects (bonds, angles, etc.) are assigned to the first matching parameter set in the XML file. For example, the following two dihedral definitions would be considered matches to the C-C-C-C dihedral in OPLS-atomtyped butane (where dummy parameters are used for the second definition):
```
<Proper class1="CT" class2="CT" class3="CT" class4="CT" c0="2.9288" c1="-1.4644" c2="0.2092" c3="-1.6736" c4="0.0" c5="0.0"/>
<Proper class1="CT" class2="CT" class3="CT" type4="opls_135" c0="3.0" c1="0.0" c2="0.0" c3="0.0" c4="0.0" c5="0.0"/>
```
However, if these parameters were found in an XML used by Foyer for atomtyping and parametrization, the parameters for the first dihedral definition would be assigned to all C-C-C-C dihedrals as this is the *first* matching dihedral. This is due to the way OpenMM assigns parameters (this is expected to be addressed in the future, see pandegroup/openmm#1959).

This PR solves this problem by adding an XML preprocessing step where topology object definitions are sorted by the number of `type` attributes defined. In the above example, the second definition would be assigned to C-C-C-C dihedrals where one of the end atoms was of type "opls_135". This method of sorting was chosen as the presence of `type` attributes indicates specificity, with more `type` attributes indicating a more specific parameter set. The order of the parameters is otherwise left unchanged, so if two matching definitions exist with the same number of `type` attributes (where Foyer has no way of distinguishing the desired precedence) then the expected behavior from OpenMM of choosing the first matching definition occurs.

These changes resolve #135.


summeraz:
It looks like invalid XML files (such as lacking a closing tag for an element) can cause ElementTree to break ([at this line](https://github.com/summeraz/foyer/blob/75322e41fc48bde859fd41165290d4f28d3c85e3/foyer/forcefield.py#L56)) with an error message the user might not be easily able to interpret. In this case we want the validator to catch this mistake and provide the user with a more descriptive error message. I've gone ahead and simply moved the topology sorting into a try block, where if an error occurs the user is provided with a warning that auto-sorting could not be performed due to an XML problem and the constructor continues on to the validator which should provide a more descriptive error message.

foyer/forcefield.py
```diff
@@ -51,6 +52,34 @@ def preprocess_forcefield_files(forcefield_files=None):
...
+                    element[:] = sorted(element, key=lambda child: -1 * len([attr_name for attr_name in child.keys() if 'type' in attr_name]))
+            xml_contents = ET.tostring(root, method='xml')
+            '''
+            Convert the output byte string to unicode. This method is specific to
+            the Python version (2 or 3)
+            '''
+            try:
+                xml_contents = unicode(xml_contents, 'utf-8')
+            except NameError:
+                xml_contents = str(xml_contents, 'utf-8')
```
ctk3b: I think you can get rid of this and simply write `xml_contents = ET.tostring(root, method='xml').decode()`

foyer/forcefield.py
```diff
@@ -51,6 +52,34 @@ def preprocess_forcefield_files(forcefield_files=None):
...
+        try:
+            '''
+            Sort topology objects by precedence, defined by the number of
+            `type` attributes specified, where a `type` attribute indicates
+            increased specificity as opposed to use of `class`
+            '''
+            root = ET.fromstring(xml_contents)
+            for element in root:
+                if 'Force' in element.tag:
+                    element[:] = sorted(element, key=lambda child: -1 * len([attr_name for attr_name in child.keys() if 'type' in attr_name]))
```
ctk3b: Nice! Could you break up the line?

summeraz: I wasn't sure exactly what the best way was to go about this. I've modified it, but let me know if there's a solution that you think is more readable.

summeraz:
I've added a section to the documentation detailing how parameters should be defined and how they are assigned. Let me know if you think this needs any modifications.

I think I'm going to add a few more unit tests (e.g. ensuring that parameter assignment is still based on order of definition within the XML if two matching parameter definitions have the same number of `type` definitions), and then I this should be good to go.

ctk3b:
Sweeeeet. LGTM! Thanks @summeraz

Issue #71, opened 2017-02-20
Optional write XML of "used" types


summeraz:
Provide an option (possibly as an optional argument to forcefield.apply()) to write an XML file containing only the information that was used in atom typing a particular system.  This could also maybe be handled via a structure.write_xml() function.  This would provide an easy-to-read format for checking to make sure the atomtyping was performed correctly.

PR #161, opened 2018-03-19
Generate a Foyer XML from a parameterized ParmEd Structure

summeraz:
This PR introduces a new method to the ParmEd `Structure`, `write_foyer`, that allows a parameterized `Structure` to be written to a Foyer XML. This addresses #71, and should be useful for archiving/documenting the specific force field parameters used for a particular system (particularly if parameters were obtained from a large Foyer XML such as OPLS). This could also prove useful if Foyer XMLs are used by simulation engines such as HOOMD to read force field parameters (so that these do not need to be manually specified by the user).

I wasn't sure exactly where this function best fit into the code base. I think having this be a method of `Structure` makes sense, rather than as a method of `Forcefield` and requiring the user to pass a parameterized `Structure` as an argument. In this case we could of course file a PR to ParmEd, but I like the idea of keeping this within Foyer so we can more easily make modifications if/when needed. The function does have the option to pass a Foyer `Forcefield` as an argument, as this provides access to additional information for the `AtomTypes` section (e.g. atom classes, SMARTS definitions) that are not stored within `Structure`. Another optional argument, `unique`, is available that allows the user to specify whether only unique atom, bond, etc. definitions should be written to the XML, or if a definition should be written for each atom, bond, etc. instance within a system (more on this later). If a `Forcefield` is passed to this function the resulting XML should be able to successfully re-atomtype the system, yielding the exact same parameterization. I will need to set up some unit tests to ensure this is actually the case, but wanted to go ahead and get the PR started.

Back to the `unique` argument. Another, and perhaps more significant, use-case for the `write_foyer` function concerns the testing of Foyer XMLs for correct parameterization (this is based on a conversation that @chrisiacovella, @justinGilmer, and myself had last week). Currently we are testing Foyer XMLs on "test" molecules by comparing atomtypes assigned by the force field to those contained within a MOL2 or Gromacs TOP file that has been provided by the user (i.e. these contain the known, correct types). However, we do not have a good way to test if bonds, angles, etc. are being properly assigned. I believe a solution to this problem is to compare Foyer XMLs in an expanded format that contains definitions for each atom, bond, etc. within a system (thus describing the full system topology, as would be found within a Gromacs TOP file). This can be achieved with `write_foyer` by providing `unique=False`, which, for example, creates a `NonbondedForce` definition for each atom (where `id` is added as an additional attribute representing the atom ID) :
`<Atom id="0" type="opls_135" charge="-0.18" sigma="3.5" epsilon="0.276144"/>`
and, e.g., creates a `HarmonicBondForce` definition for each bond that contains the appropriate atom IDs:
`<Bond id1="0" id2="3" type1="opls_135" type2="opls_140" length="0.109" k="284512.0"/>`.
If we replace the MOL2 and TOP files we are currently using for test molecules with these expanded Foyer XMLs, we can easily parse these with [ElementTree](https://docs.python.org/2/library/xml.etree.elementtree.html) and compare not only atomtype information, but all bond, angle, etc. parameterization as well.

Usage of `write_foyer` is pretty straightforward. Here is an example where a Foyer XML is written for an ethane molecule created with mBuild:
```
ethane = Alkane(2)
opls = Forcefield(name='oplsaa')
ethane_a = opls.apply(ethane)
ethane_a.write_foyer('from-structure.xml', forcefield=opls)
```
and here is the resulting XML:
```
<ForceField>
  <AtomTypes>
    <Type name="opls_135" class="CT" element="C" mass="12.01078" def="[C;X4](C)(H)(H)H" desc="" doi="10.1021/ja9621760"/>
    <Type name="opls_140" class="HC" element="H" mass="1.007947" def="H[C;X4]" desc="" doi="10.1021/ja9621760"/>
  </AtomTypes>
  <NonbondedForce>
    <Atom type="opls_135" charge="-0.18" sigma="3.5" epsilon="0.276144"/>
    <Atom type="opls_140" charge="0.06" sigma="2.5" epsilon="0.12552"/>
  </NonbondedForce>
  <HarmonicBondForce>
    <Bond type1="opls_135" type2="opls_135" length="0.1529" k="224262.4"/>
    <Bond type1="opls_135" type2="opls_140" length="0.109" k="284512.0"/>
  </HarmonicBondForce>
  <HarmonicAngleForce>
    <Angle type1="opls_135" type2="opls_135" type3="opls_140" angle="1.932079482" k="313.8"/>
    <Angle type1="opls_140" type2="opls_135" type3="opls_140" angle="1.8814649337" k="276.144"/>
  </HarmonicAngleForce>
  <RBTorsionForce>
    <Proper type1="opls_140" type2="opls_135" type3="opls_135" type4="opls_140" c0="0.6276" c1="1.8828" c2="0.0" c3="-2.5104" c4="0.0" c5="0.0"/>
  </RBTorsionForce>
</ForceField>
```

A few other things to note:

- As shown in the above XML, I don't think we are parsing the `desc` argument properly. I am grabbing this directly from `Forcefield.atomTypeDesc` and this always seems to be an empty string.
- An additional argument should be added to `apply` (and also to `mBuild.Compound.save`) to allow a Foyer XML to be written at the time the forcefield is applied.
- I have a feeling [this block of code](https://github.com/summeraz/foyer/blob/4d4c7655cdacda0d6e6c42559fc450aa742c0b08/foyer/xml_writer.py#L74-L98) could be written in a more Pythonic manner, but I've struggled to find a better solution.
- We need to properly document the rounding that is used for each parameter. I basically just used rounding based on the parameters we have defined in the OPLS forcefield XML.
- I need to add `coulomb14scale` and `lj14scale` attributes to `NonbondedForce`. These are stored somewhere within the parameterized `Structure`, but their location is not obvious...
- `write_foyer` currently considers bond, angle, etc. definitions to be unique if the tuple of involved atom types is unique. The resulting XMLs could be condensed if we grouped parameters using atom classes. Since we have access to the `class` for each atomtype if a `Forcefield` is passed using the `forcefield` argument, I think that grouping parameters by class is something that should be doable.


davidlmobley:
Big picture question: If you guys are working with ParmEd structures, is this a potential entrance route for other force fields into foyer? e.g. can a "port" of a force field or subset of a force field be made into Foyer using ParmEd? Or are you doing something different here?

We should probably have another call with you guys soon. We now have a working SMARTS/SMIRKS-based force field that basically covers all drug-like chemical space with very small exceptions (more coverage than GAFF and parm@Frosst) that we've validated fairly extensively; I imagine it would be easy for you to get this into the Foyer format. And work you are doing with Foyer may also connect up with what you are doing.

If you'd like I can provide a link to a preprint very soon (hopefully it will be up today/tomorrow).

summeraz:
@davidlmobley Yes, this function should serve a few purposes, but one of those certainly will be aiding users with porting their force fields into Foyer's format (via ParmEd). From a parameterized ParmEd structure, the user should be able to use `write_foyer` to produce a complete Foyer XML that is missing only a few attributes in the AtomTypes section (e.g. `def`, `desc`, `class`), which would need to be addressed manually (we've discussed adding a routine into Foyer to "guess" SMARTS strings, so if/when that is implemented it could be merged with this function as well).

Of course the primary hurdle here is obtaining the parameterized ParmEd structure, but that should not be a problem in the case of your force field, so yes with the help of this function we should hopefully be able to pretty easily convert your force field into Foyer's format, and it would be a welcome addition. If you don't mind providing the preprint link once it's available I would definitely be interested in checking it out. A call in the near future to further discuss this certainly seems like a good idea.

davidlmobley:
Preprint is up! https://doi.org/10.1101/286542 Please feel free to share; happy to discuss when you guys are ready.

davidlmobley:
And for a one-stop quick peek inside, see https://github.com/openforcefield/smirnoff99Frosst/blob/master/smirnoff99Frosst/smirnoff99Frosst.offxml

mattwthompson:
Sometimes we use `parmed` classes that inherit from `pmd.Structure`. As I understand their use, it is preferred to stay in the other classes and not convert back to `pmd.Structure` (actually, I'm not even sure you can). To get this function to work with those classes, we would need to add this function to each of them

 ```
pmd.Structure.write_foyer = write_foyer
pmd.amber._amberparm.AmberParm = write_foyer
pmd.gromacs.GromacsTopologyFile = write_foyer
```

and probably more. This seems not-so-preferred. But I'm not sure a better approach.

I encountered this when trying to convert the force field information in a `.prmtop` file to a foyer XML.

**Comment by ctk3b** [comment]
One way to get around this technical limitation is to simply write this function to accept a `pmd.Structure` instead of implementing it as a method.

**Comment by mattwthompson** [comment]
Anything blocking this from being merged?

P.S. this is an interesting example of adding a feature to parmed temporarily while we wait for `topology.core` to be developed enough for big additions

**Comment by mattwthompson** [comment]
Reviving this because I think it's cool.

Looks like it's not too far away - there are the changes I needed to make to the written file to get it to be read back in (as @summeraz noted in the docstring):

```diff
3,4c3,4
<     <Type name="opls_135" class="CT" element="C" mass="12.01078" def="[C;X4](C)(H)(H)H" desc="" doi="{'10.1021/ja9621760'}"/>
<     <Type name="opls_140" class="HC" element="H" mass="1.007947" def="H[C;X4]" desc="" doi="{'10.1021/ja9621760'}"/>
---
>     <Type name="opls_135" class="" element="C" mass="12.01078" def="[C;X4](C)(H)(H)H" desc="" doi=""/>
>     <Type name="opls_140" class="" element="H" mass="1.007947" def="H[C;X4]" desc="" doi=""/>
6c6
<   <NonbondedForce>
---
>   <NonbondedForce coulomb14scale="0.5" lj14scale="0.5">
```

mattwthompson:
This is still limited in scope but I have attempted to write the `<NonbondedForce coulomb14scale="0.5" lj14scale="0.5">` part, including values generated by parsing the `structure.adjusts`. The test could be beefed up plenty - maybe @ahy3nz knows how to get these two numbers from the `Forcefield` object?

If this addition is cleared I consider this good to go.

ahy3nz:
Oh cool I didn't know about `structure.adjusts`!
In the par writer PR, we've been looking at the various dihedral.type.scnb [like here](https://github.com/mosdef-hub/mbuild/pull/508/files#diff-5d0a1b3fdc31ba492a31680b233c6f12R72). The idea is that every `structure.dihedral.type` or `structure.rb_torsion.type` has a `scee` and `scnb` property (so the `pmd.structure` could have arbitrary 1-4 QQ and LJ scalings for every 1-4 interaction). 

I think the logic would follow similarly to how you iterate through all the `adjusts` and hope for a single `chgscale` value; iterate through all the `rb_torsions.type` or `dihedral.type` and hope for a single `scnb` value. I think `scee = 1/chgscale` so the 14 LJ scaling is `1/scnb`?

mattwthompson:
I would think (?) that these approaches should produce the same value

mattwthompson:
I don't know if this is concerning but they don't match, at least in the structure used in the unit test. But they are the same before and after the loop of writing a partial XML, reading it back in, and typing with it again

```python3
(Pdb) [t.type.scee for t in typed.rb_torsions]
[1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
(Pdb) [t.type.scee for t in typed_by_partial.rb_torsions]
[1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
```
```python3
(Pdb) [t.type.epsilon/((t.atom1.atom_type.epsilon * t.atom2.atom_type.epsilon) ** 0.5) for t in typed_by_partial.adjusts]
[0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5]
(Pdb) [t.type.epsilon/((t.atom1.atom_type.epsilon * t.atom2.atom_type.epsilon) ** 0.5) for t in typed.adjusts]
[0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5]
```

ahy3nz:
Hmm, I think parmed has multiple ways of handling the 14s. In the par_writer PR and examples, we were just using parmed to load in top/gro file. The gromacs/charmm readers in parmed will store these 14 interaction parameters in `dihedral.scee` or `dihedral.scnb`. 

In the foyer workflow, we're dealing with `omm.System` and then `pmd.Structure`. In `omm.System` the 14 interactions are stored in `omm.NonbondedForce` as `NonbondedExceptions` to the 14 interactions. When we convert `omm` > `pmd`, the `omm.NonbondedExceptions` appear to get stored as `pmd.structure.adjusts`

So it looks like there are multiple ways of handling these 14s, `scee/scnb` for charmm/gromacs, `adjusts` for openmm; and it doesn't look like they'll mutually update each other.

That being said, it's probably better to look for `structure.adjusts` and not expect agreement with `structure.dihedral.scee/scnb`. So I think your current implementation is more appropriate - the only  way (i think) to get the actual lj14scale or coulomb14scale attribute would involve [getting the information from the `omm.NonbondedGenerator`](https://github.com/pandegroup/openmm/blob/25e12f36471c7224fad948a8384b974f14a14db6/wrappers/python/simtk/openmm/app/forcefield.py#L2328)

mattwthompson:
Looks like the coverage reports are complaining because a) this project has relatively few lines of code so any change has a large impact on the overall coverage and b) the unit test looks for RB torsions and not proper dihedrals. I can add another molecule if people want but I'm generally fine with this as it is.

https://coveralls.io/builds/23222238/source?filename=foyer%2Fxml_writer.py#L50

I'm also not super inclined to dig through the OpenMM objects because I'm not familiar with them and we're going to be ditching them in short time. If somebody else wants to implement this please do so. Again, open to objections but I'm happy with where this is at the moment.

mattwthompson:
@justinGilmer I'd like to fix or at least investigate the issues you found in #237 before merging. Should be able to cover a lot with just another test molecule or two.

mattwthompson:
@ahy3nz @justinGilmer I think this is good to go but please give these changes a once-over and then we can merge.

mattwthompson:
Added a test for writing `PeriodicTorsionForce` to make codecov happy and it's good I did, there was a small bug in how a variable was stored. Coverage still drops but only by a small amount.

foyer/xml_writer.py
```diff
@@ -0,0 +1,281 @@
...
+                if atypes[0] > atypes[-1]:
+                    atypes = atypes[::-1]
+            else:
+                dihedral_force.set('id1', str(dihedral.atom1.idx))
+                dihedral_force.set('id2', str(dihedral.atom2.idx))
+                dihedral_force.set('id3', str(dihedral.atom3.idx))
+                dihedral_force.set('id4', str(dihedral.atom4.idx))
+        for id in range(4):
+            dihedral_force.set('type{}'.format(id+1), atypes[id])
+        dihedral_force.set('periodicity1', str(dihedral.type.per))
```
ahy3nz: Okay this is where things get hairy. In these periodic torsions (like rbtorsions), you can essentially have multiple terms (sum of a series). This chunk of code is only looking at one term within the series, when sometimes there are multiple (like line 96493 in  [here](https://github.com/choderalab/openmm-forcefields/blob/master/charmm/ffxml/charmm36_nowaters.xml))

UNLIKE rbtorsions, this series does not get lumped into a single `structure.dihedral` term like it does in `structure.rb_torsions`. If there's a periodictorsion with multiple periodicities in the XML element, that gets "unwrapped" into multiple `structure.dihedral` terms (it looks like they're still adjacent elements in the `structure.dihedrals` list, though).

So I think there needs to be some sort of check for multiple periodicities by looking at adjacent terms. Sometimes there will be, sometimes there won't be. If there ARE multiple periodicities, then they should all get set in the same `periodictorsion` xml element.

The next problem (I need to investigate this further), is that the parameterized-structures all have `dihed.funct` == '1', which is okay for single-term periodictorsions, but should be `dihed.funct` == '9' for multi-term periodictorsions (I think, doing some comparisons with some of my charmm phospholipid parameters)

ahy3nz: That last problem with the `dihed.funct` might be a parmed issue, but i'll get back to this

foyer/xml_writer.py
```diff
@@ -0,0 +1,281 @@
...
+        if not unique and child.tag != 'AtomTypes':
+            continue
+        child[:] = sorted(child, key=lambda elem:
+            tuple(elem.get(id) for id in sortby[child.tag]))
+        elems_to_remove = []
+        for elem in child:
+            if _elements_equal(elem, prev):
+                elems_to_remove.append(elem)
+                continue
+            prev = elem
```
ahy3nz: Okay so if I'm interpreting this right, we're iterating through the XML elements and looking for match between the previous line and the current line. A match is defined as matching everything (like for a dihedral, type1/2/3/4, constants, ref values, periodicites etc). I think this works 99% of the time when each XML element refers to some unique bonded-tuple, and each XML element fully defines the bonded interaction for that bonded-tuple

Going back to the multiple-periodic torsion issue from above, we end up breaking this logic. In the ethane example, H-C-C-H could have multiple periodicities (let's say 2 periodicities). When this happens, the raw XML (still containing duplicates) will contain alternating XML elements with HCCH parameters for periodicity1, then HCCH parameters for periodicity2. Because we are just comparing previous and current XML elements, we effectively never compress the unique dihedrals. 

This seems to be a symptom/consequence of the earlier periodic torsion issues I outlined earlier - if the original issue gets addressed, I think this will get cured as well

mattwthompson:
> I say fix the sigma-off-by-10 issue and merge the PR

That sounds like a plan to me; remember that Andrew wrote most of this code so I'd be more comfortable with you leading these changes and me following along. Because yes, that is quite a mess.

foyer/xml_writer.py
```diff
@@ -0,0 +1,281 @@
...
+    nonbonded.set('coulomb14scale', str(_infer_coulomb14scale(self)))
+    nonbonded.set('lj14scale', str(_infer_lj14scale(self)))
+    atomtype_attrs = collections.OrderedDict([
+        ('name', 'name'),
+        ('class', 'forcefield.atomTypeClasses[name]'),
+        ('element', 'forcefield.atomTypeElements[name]'),
+        ('mass', 'atom.mass'),
+        ('def', 'forcefield.atomTypeDefinitions[name]'),
+        ('desc', 'forcefield.atomTypeDesc[name]'),
+        ('doi', 'forcefield.atomTypeRefs[name]')
```
ahy3nz: Doing the iteration through all these XML attributes, the resultant XML looks like
`<Type name="opls_135" class="CT" element="C" mass="12.01078" def="[C;X4](C)(H)(    H)H" desc="" doi="{&quot;{'10.1021/ja9621760'}&quot;}"/>`

When it should be 
` <Type name="opls_135" class="CT" element="C" mass="12.01078" def="[C;X4](C)(H)(    H)H" desc="" doi="{'10.1021/ja9621760'}"/>`

There's some sort of weird formatting for the doi, I wonder if this can be done by stripping out the `{&quot;` from the doi attribute

PR #162, opened 2018-03-19
WIP: Foyer webpage

justinGilmer:
Provdided the basic RTD format that `mBuild` uses on its
github.io webpage.

Converted current markdown files to RST.

Added travis directives to generate these pages when a successful test
passes, like current `mBuild` functionality.

Modifications still need to be completed to make the site a bit more full-fledged.

Add more documentation for the actual data structures, etc. Adding that next.

Also document features of `foyer` like the residue templating, etc.


justinGilmer:
Not ready for actual merging, just putting this here so others can pull and generate the HTML pages themselves as more changes are made.

ctk3b:
Nice thanks for starting on this @justinGilmer. 

Not sure if it's worth the time right now since this works just fine but I would like to try moving back to readthedocs at some point so we can remove the doc deploy process that we currently use.

justinGilmer:
Maybe I'm misunderstanding, but what is the difference between that and what we use now? I assumed RTD is just a theme for generating a website? Or does it contain more than that?

ctk3b:
They do have a sphinx them but they're also a service: https://readthedocs.org/

Every commit to master/tags/whatever-you-configure will create a new version of the docs and do all the sphinx build behind the scenes for you. We used to use it but ran into a lot of issues related to package installation. I think most of those have been resolved at this point but haven't fully re-evaluated

summeraz:
We can obviously continue to improve this, but I think the website is at a point now where it's good to go.

mattwthompson:
Am I able to preview the site (at a webpage) before merging?

summeraz:
@mattwthompson Yes, you’ll have to [add justinGilmer as a remote and fetch this branch.](https://stackoverflow.com/a/9153737). Then navigate to the `docs` directory, run `make html` to generate the html files, and view the website with `open _build/html/index.html`.

Issue #164, opened 2018-05-02
CustomTorsionForce Elements in FFXML


ahy3nz:
The Charmm forcefield includes elements for improper dihedrals (which are supported in OpenMM and ParmEd). Foyer needs to accept CustomTorsionForce elements also

ahy3nz:
Charmm XML that was designed for OpenMM (https://github.com/choderalab/openmm-forcefields)

PR #165, opened 2018-05-02
Charmm torsions allowable in forcefield XML files

ahy3nz:
XML schema needs to account for `CustomTorsionForce` #164


ahy3nz:
As of this latest commit, ff.xsd includes CustomTorsionForce (impropers) and AmoebaUreyBradlyForce (urey bradley terms). Interestingly, the urey bradley terms are just specified as bonds and bondtypes in the parmed structure (with the correct UB parameters though). This could have implications for the mbuild lammpsdata format, since the charmm angle_style requires the eq angle, angle constant, ub distance, and ub force constant. But this charmm angle style could also be split into harmonic angles and harmonic bonds terms

foyer/forcefields/ff.xsd
```diff
@@ -52,6 +52,20 @@
       </xs:simpleContent>
     </xs:complexType>
   </xs:element>
+  <xs:element name="UreyBradley">
+    <xs:complexType>
+      <xs:simpleContent>
+        <xs:extension base="xs:string">
+          <xs:attribute type="xs:float" name="d" use="optional"/>
+          <xs:attribute type="xs:float" name="k" use="optional"/>
+          <xs:attribute type="xs:string" name="type1" use="optional"/>
```
summeraz: Do you know if OpenMM supports the use of "class" here in addition to "type"? I'm guessing they do, since this seems to be the case for all of the other force definitions. I'm sure they have this documented somewhere although I'm having trouble finding any information...

foyer/forcefields/ff.xsd
```diff
@@ -52,6 +52,20 @@
...
+          <xs:attribute type="xs:string" name="type2" use="optional"/>
+          <xs:attribute type="xs:string" name="type3" use="optional"/>
+        </xs:extension>
+      </xs:simpleContent>
+    </xs:complexType>
+  </xs:element>
+
```
summeraz: Extraneous empty line

foyer/forcefields/ff.xsd
```diff
@@ -185,6 +222,16 @@
...
     </xs:complexType>
   </xs:element>
+  <xs:element name="CustomTorsionForce">
+    <xs:complexType>
+      <xs:choice maxOccurs="unbounded">
+        <xs:element ref="Proper" minOccurs="0"/>
+        <xs:element ref="PerTorsionParameter" minOccurs="0"/>
+        <xs:element ref="Improper" minOccurs="0"/>
+      </xs:choice>
+      <xs:attribute type="xs:string" name="energy" use="optional"/>
```
summeraz: Does `CustomTorsionForce` not require an energy definition?

ahy3nz:
Any fixes to #167  will depend on foyer accepting charmm force fields, so I'll tack on the sorting of 1-2 and 1-3 bonds in this PR too

ahy3nz:
Latest commit includes a function to split up the HarmonicBondForce into 2 HarmonicBondForce objects (1 for harmonic bonds, the other for 1-3 UB bonds). The idea is to find the HarmonicBondForce object, compare the atoms in each bond to the topology's bond network, and copy the bond parameters into either the new harmonic bond or 1-3 UB bond object. 

I wasn't able to find a way to simply remove bond parameters from a Force object, so I just made 2 new Force objects and removed the old Force object from the system. 

This update works with my small, dihedral system (no false impropers, no false bonds)

foyer/forcefield.py
```diff
@@ -187,6 +187,34 @@ def _update_atomtypes(unatomtyped_topology, res_name, prototype):
             for old_atom, new_atom_id in zip([atom for atom in res.atoms()], [atom.id for atom in prototype.atoms()]):
                 old_atom.id = new_atom_id
 
+def _resolve_urey_bradleys(system, topology):
```
ctk3b: Would this be better named `_separate_urey_bradley_bonds`?

foyer/forcefield.py
```diff
@@ -405,6 +434,10 @@ def apply(self, topology, references_file=None, use_residue_map=True,
 
         return structure
 
+    
+
+
+
```
ctk3b: extra whitespace

foyer/forcefields/ff.xsd
ctk3b: I suspect so but would be good to confirm. If so, the schema should allow this.

foyer/forcefields/ff.xsd
ahy3nz: Yes, looks like it's a required attribute

foyer/forcefield.py
```diff
@@ -210,6 +210,34 @@ def _update_atomtypes(unatomtyped_topology, res_name, prototype):
...
+        system : openmm System
+
+        """
+        atoms = [a for a in topology.atoms()]
+        bonds = [b for b in topology.bonds()]
+        ub_force = mm.HarmonicBondForce()
+        harmonic_bond_force = mm.HarmonicBondForce()
+        for i, force in enumerate(system.getForces()):
+            if isinstance(force, mm.HarmonicBondForce):
+                for j in range(force.getNumBonds()):
```
ctk3b: nit: could we name these something like `bond_idx` and `force_idx`?

foyer/forcefield.py
```diff
@@ -210,6 +210,34 @@ def _update_atomtypes(unatomtyped_topology, res_name, prototype):
...
+                    if (atoms[force.getBondParameters(j)[0]], 
+                        atoms[force.getBondParameters(j)[1]]) not in bonds:
```
ctk3b: Is the ordering of atoms within a bond guaranteed to be the same in both objects?

summeraz:
I think all this needs is some unit tests for proper urey-bradley and custom torsion assignment and it should be good to go. [This test](https://github.com/mosdef-hub/foyer/blob/83a3c56d680b82822c5cfafd415d2d35bc7297ff/foyer/tests/test_forcefield.py#L106) should serve as a good example.

rmatsum836:
I'm looking through this in greater detail now, but the tests are passing for me locally.

mattwthompson:
Appveyor is failing due to an unrelated network failure. Maybe @justinGilmer can restart it, I can't.

ahy3nz:
PR Summary:

* Expand xml schema to include `PeriodicTorsionForce`(Charmm dihedrals) and `CustomTorsionForce` (Charmm impropers) and `AmoebaUreyBradleyForce` (Charmm UB terms, harmonic angle + 13 bond)
* Create 2 `HarmonicBondForce` objects for the ommSystem (one for harmonic bonds, the other for urey bradley bonds) such that conversion to pmdStructure will populate `structure.bonds` and `structure.urey_bradleys`
* The resultant pmdStructure should be able to process/understand all these omm objects

PR #168, opened 2018-05-30
Multiple DOIs per atomtype

summeraz:
Currently the `doi` attribute for atomtypes defined in a force field XML file only supports a single DOI. However, there may be instances where multiple DOIs would be appropriate (see #158). This PR adds support for listing multiple DOIs within the `doi` attribute, and passes this information through to the BibTeX file that is created.

### DIFF
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -723,9 +723,9 @@ class Forcefield(app.ForceField):
                 warnings.warn("Reference not found for atom type '{}'."
                               "".format(atype))
         unique_references = collections.defaultdict(list)
-        for key, values in atomtype_references.items():
-            for doi in values:
-                unique_references[doi].append(key)
+        for atomtype, dois in atomtype_references.items():
+            for doi in dois:
+                unique_references[doi].append(atomtype)
         unique_references = collections.OrderedDict(sorted(unique_references.items()))
         with open(references_file, 'w') as f:
             for doi, atomtypes in unique_references.items():
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -335,8 +335,7 @@ class Forcefield(app.ForceField):
         if 'des' in parameters:
             self.atomTypeDesc[name] = parameters['desc']
         if 'doi' in parameters:
-            dois = [doi.replace(',', '').strip() for doi
-                    in parameters['doi'].split(',')]
+            dois = set(doi.strip() for doi in parameters['doi'].split(','))
             self.atomTypeRefs[name] = dois
 
     def apply(self, topology, references_file=None, use_residue_map=True,
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -335,7 +335,9 @@ class Forcefield(app.ForceField):
         if 'des' in parameters:
             self.atomTypeDesc[name] = parameters['desc']
         if 'doi' in parameters:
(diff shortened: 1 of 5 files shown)


foyer/forcefield.py
```diff
@@ -721,14 +723,16 @@ def _write_references_to_file(self, atom_types, references_file):
                 warnings.warn("Reference not found for atom type '{}'."
                               "".format(atype))
         unique_references = collections.defaultdict(list)
-        for key, value in atomtype_references.items():
-            unique_references[value].append(key)
+        for key, values in atomtype_references.items():
```
ctk3b: Could we update these to be `for atomtype, dois in ...`?

foyer/tests/test_forcefield.py
```diff
@@ -74,6 +75,17 @@ def test_write_refs():
...
 
+def test_write_refs_multiple():
+    mol2 = mb.load(get_fn('ethane.mol2'))
+    oplsaa = Forcefield(forcefield_files=get_fn('refs-multi.xml'))
+    ethane = oplsaa.apply(mol2, references_file='ethane-multi.bib')
+    assert os.path.isfile('ethane-multi.bib')
+    with open(get_fn('ethane-multi.bib')) as file1:
+        with open('ethane-multi.bib') as file2:
+            diff = difflib.ndiff(file1.readlines(), file2.readlines())
+        changes = [l for l in diff if l.startswith('+ ') or l.startswith('- ')]
```
ctk3b: Check out https://docs.python.org/3/library/filecmp.html

summeraz: So I had initially tried using `filecmp` when writing [this test](https://github.com/mosdef-hub/mbuild/blob/1468900d2121661947b221a8f2b5534405ca9fcc/mbuild/tests/test_utils.py#L15-L22) for mBuild; however, I ran into an issue with Appveyor (https://ci.appveyor.com/project/ctk3b/mbuild-o0viu/build/1.0.755#L1137). I think because `filecmp` performs a binary comparison there is an issue that arises when comparing files generated on MacOS/Linux and on Windows (see https://bugs.python.org/issue6306).

foyer/forcefield.py
summeraz: 👍

foyer/tests/test_forcefield.py
ctk3b: Ah I see - makes sense. If there is no diff, couldn't you simply `assert not diff`? Or will it always contain something? Is it just a header that you can unambiguously strip?

Regardless of the above, I would de-dent the `changes = ...` and `assert` lines.

summeraz: `diff` will show all lines, with a `+ ` in front of lines that are unique to `file2` and a `- ` in front of lines that are unique to `file1`. So `changes` is just stripping out only those lines that are different between the two files. I actually got this from [this StackOverflow answer](https://stackoverflow.com/questions/15864641/python-difflib-comparing-files/15864920#15864920).

ctk3b: Heh alright this isn't really that essential but one last suggestion: https://docs.python.org/2/library/difflib.html#difflib.unified_diff and just set `n=0` to get no context lines! 🎉

summeraz: 🍾 Ah, nice suggestion! It seemed like there had to be a better way to compare two files.

PR #171, opened 2018-06-12
Replace ring finding code

### LINKED ISSUE #170: Incorrect atom typing?
You will need to remove the extra .txt on these files.
[pc71bm.hoomdxml.txt](https://github.com/mosdef-hub/foyer/files/2073145/pc71bm.hoomdxml.txt)
[opv_ff.xml.txt](https://github.com/mosdef-hub/foyer/files/2073146/opv_ff.xml.txt)
 
```
pc71bm = mb.load("pc71bm.hoomdxml")
foyer_kwargs = {"assert_angle_params": False,
                "assert_dihedral_params": False}
pc71bm.save("pcbm_box_other.hoomdxml", forcefield_files="opv_ff.xml", overwrite=True, foyer_kwargs=foyer_kwargs)
```
The C atoms in the rings should be CA, and the other C atoms should be typed CT.

summeraz:
Our previous approach to finding the rings that each atom is included within, [which relied upon NetworkX's cycle_basis function](https://github.com/mosdef-hub/foyer/blob/6b61e9de709d43ddad0f63e06543f5a4b3f91f38/foyer/smarts_graph.py#L259), did not guarantee that all rings would be found for multi-cyclic systems. This function would return a set of rings from which the full bond graph could be recreated, but this would not necessarily be all rings in the system.

This PR replaces `cycle_basis` with a function that should find all (chordless) cycles (i.e. rings) that each atom is included within, loosely based off of [this StackOverflow post](https://stackoverflow.com/a/4028855).

Resolves #170


summeraz:
This is still a WIP. I need to set up some unit tests and check to see how much of a performance hit this takes. The code included now was just the first successful attempt, so I'm sure there are plenty of optimizations that can be made.

@mikemhenry I tested this code locally on your system (from #170) and all atom types were assigned correctly, so you can checkout this branch if you need to atom type similar systems in the meantime before this is merged.

mikemhenry:
This is great, works for me and will be REALLY useful in other compounds I'm working with!

mikemhenry:
If pytest supports it, also running the test like 3 times to make sure
things stay deterministic would be good to. Also an assertion that the atom
types are correct would be good to.

foyer/tests/test_graph.py
```diff
@@ -37,3 +37,13 @@ def test_lazy_cycle_finding():
...
         assert all([hasattr(a, 'cycles') for a in top.atoms()])
+
+
+def test_cycle_finding_multiple():
+    fullerene = pmd.load_file(get_fn('fullerene.pdb'), structure=True)
+    top, _ = generate_topology(fullerene)
+
+    _prepare_atoms(top, compute_cycles=True)
+    cycle_lengths = [list(map(len, atom.cycles)) for atom in top.atoms()]
+    expected = [5, 6, 6]
```
ctk3b: I might be thinking about this incorrectly but isn't this an entire fullerene structure? How can there only be 3 cycles?

ctk3b: Ah! Nice - LGTM

summeraz:
Performance tests have been added. I'm not sure we actually need additional tests for correct atom-typing and deterministic behavior since we already test those aspects with `test_opls.py` and `test_trappe.py` and [this test](https://github.com/summeraz/foyer/blob/57580a39bf9e2ebc7ea4fdb5e217c51f9645d280/foyer/tests/test_graph.py#L42-L49) ensures that the ring-finding is deterministic.

Issue #140, opened 2017-08-25
Support for purely non-bonded forcefields


mattwthompson:
This might not be feasible, but I thought I'd document in just in case.

Some forcefields rely purely on non-bonded interactions, even for interactions that are effectively bonded, i.e. the Si-O bond in silica. One such example is [ClayFF](http://pubs.acs.org/doi/abs/10.1021/jp0363287), and I'm sure there are others for silica and various covalently-bonded solids out there. The obvious issue is that no SMARTS string can be generated when the chemical topology is not defined by bonds. A workaround I've been using is generating bonds for atomtyping and removing them after atomtyping, but this is not a good general solution.

summeraz:
The BKS force field for silica takes a similar approach. In this force field, however, there is one set of parameters for all silicon atoms and one set for all oxygen atoms. So the SMARTS strings would just be [Si;X0] and [O;X0]. I'm guessing the force field you want to use has several atom-types defined for a single element type? If so, then how are these atom-types distinguished in the force field?

mattwthompson:
The [O;X0] approach handles a lot of cases. But take for example how ClayFF has different parameters for octahedral and tetrahedral aluminum; that wouldn't be be handled with [Al;X0] if you stumbled across a mineral with both types of aluminum.

I don't think there's a generalized way to build more complex structures with accurate atomtypes, and LAMMPS seems to be preferred by people who study clay. Clays and similar materials are solid enough that if you had two different oxygen types in the bulk phase of a solid, they wouldn't diffuse far enough away from their original position to confuse the topology.

chrisiacovella:
I ran into similar issues with InterfaceFF for mica...basically the bonded structure didn't necessarily differentiate some of the types. Creating fictitious bonds is an interesting idea. Since I was dealing a flexible crystalline structure that was basically just a unit cell I'd duplicate (and was already atom-typeD), I just give each element a unique name, and defined a SMARTS that was essentially just that name (i.e. [Al001] or [Si002]) for each atom-type.  I'm sure there was actually probably just a minimal set of atoms I would need to define (since some could be described well by bonded structure), but this wasn't very hard.

Issue #128, opened 2017-07-05
Handling of custom cross-interactions


mattwthompson:
As far as I can tell, there's no simple way to set cross-interactions that differ from what a combining rule would specify. In other words, I can't manually set the nonbonded interaction between atomtype i and atomtype j, which I would like to do for certain custom forcefields. I'm not sure exactly how openmm handles it (if it does) but [parmed appears to](https://github.com/ParmEd/ParmEd/blob/e4173b177f78b7321ae2cd9dd546349ec9e82754/parmed/gromacs/gromacstop.py)

summeraz:
It appears this is doable within an OpenMM XML (see [this example](https://github.com/choderalab/openmm-forcefields/blob/57892fd64405aa72a9980c827a0c41733afcb230/charmm/ffxml/waters_ions_spc.xml#L117)). Whether or not that information would get passed through using our current workflow is another matter...

mattwthompson:
While attempting to worth through https://github.com/mosdef-hub/mbuild/issues/455, I can say it is not supported in the current XML schema (much less passed through to the atomtyping step)

```
>>> Forcefield('nbfixpair.xml')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mwt/software/foyer/foyer/forcefield.py", line 230, in __init__
    Validator(ff_file_name, debug)
  File "/Users/mwt/software/foyer/foyer/validator.py", line 21, in __init__
    self.validate_xsd(ff_tree)
  File "/Users/mwt/software/foyer/foyer/validator.py", line 59, in validate_xsd
    xmlschema.assertValid(ff_tree)
  File "src/lxml/etree.pyx", line 3532, in lxml.etree._Validator.assertValid
lxml.etree.DocumentInvalid: Element 'LennardJonesForce': This element is not expected., line 13
```

in which the offending file (lines 13-15) are
```
  <LennardJonesForce lj14scale="1.0">
   <NBFixPair epsilon="0.01" sigma="0.02" type1="Ti3+4" type2="C_3"/>
  </LennardJonesForce>

summeraz:
[...]

PR #180, opened 2018-08-14
Support NBFix

mattwthompson:
Either a partial or complete fix for #128 (and perhaps #140), depending mostly on how this information gets moved from OpenMM to ParmEd.


ctk3b:
👍

Have you tried running either an OpenMM simulation with this or using ParmEd to write it to a different format?

mattwthompson:
Nope, and I've managed to trigger an error deep in the code while attempting to do so

mattwthompson:
Closing in favor of #267 short-term and the new backend long-term

PR #185, opened 2018-10-09
Add imidazole parameters

mattwthompson:


mattwthompson:
In 2e92c9f is that I assume that deprotonating an imidazole to an imidazolate doesn't notably affect the molecular geometry. The atom types, taken from AMBER, are very similar but some particular combinations of angles and dihedrals weren't covered.

uppittu11:
Let's update and merge this PR

PR #186, opened 2018-10-13
Add debug-level verbosity

mattwthompson:
This PR includes two small changes.

1. We've had discussions about logging some details in the atom-typing process, but nothing has happened yet. I have added a `verbose` argument to `Forcefield.apply()` that should serve as a starting point.

2. I have added the option to print out the types of the atoms in angles and dihedrals that are missing parameters. This builds on #155 - the user currently knows that *some* angles/dihedrals are missing, but this makes it easier for the user to know *which ones* are missing so that the XML or system can be fixed.


richardjgowers:
This looks cool, if we add some sort of set operation to collect all missing angles and group them by type then it (probably) won't be too spammy for just printing all the time, so no need for verbose keyword.  I can only see the verbose keyword as just being annoying because you'll run it once, fail, then run it again with verbose.

mattwthompson:
I am closing this because these commits accidentally got merged recently by an accidental branching-off that became a separate PR.

This sort of behavior is useful, though, so I'll build on it sometime in the future.

Issue #188, opened 2018-11-28
OpenMM 7.3+ breaks core functionality


mattwthompson:
Some backwards compatibility was broken when OpenMM was bumped to 7.3 and `createSystem` now completely breaks.

For now we should pin OpenMM to <7.3, but we should also work on moving the core functionality away from OpenMM and to a new, more flexible backend.

mattwthompson:
Fixed #246, thanks @ahy3nz!

PR #190, opened 2018-11-29
Pin OpenMM to < 7.3

mattwthompson:
Temporary workaround to #188

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,7 @@ environment:
 
   matrix:
     - PYTHON: "C:\\Miniconda3-x64"
+      PYTHON_VERSION: "2.7"
       CONDA_PY: "35"
       ARCH: "64"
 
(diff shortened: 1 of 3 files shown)


mattwthompson:
This fixes the openmm portion of this headache. However, conda is grabbing mdtraj 1.9.2, which has the mol2 issue. It's appropriately pinned to 1.9.1 in mbuild's [requirements.txt](https://github.com/mosdef-hub/mbuild/blob/16e9c8f024d788361494e1a5691a26f240816a51/requirements.txt) but because it's been so long since we've made an mbuild release, that change [2 months ago](https://github.com/mosdef-hub/mbuild/pull/466) is not known to conda.

This is completely analogous to the issues we were having releasing mbuild yesterday. We can't currently release foyer because of a dependence to mbuild (mdtraj) that is not a dependency of foyer, but must be pinned. And we can't build mbuild because a dependency to foyer (openmm) that is not a dependency of mbuild needs to be pinned.

We're close to fixing the issue in mdtraj and pushing a new release there, but I'm not sure when that will happen. And that wouldn't actually fix our issues, because we would still be unable to build a passing release of mbuild without these changes to foyer.

PR #193, opened 2018-12-08
simplified trappe-ua forcefield by using more broad classes

richardjgowers:
So I'm trying to add more parameters to the UA ff, but it's tedious as hell because of how restrictive the classes are.  If the class for existing atom types are broadened to be `CHx` then it is much less verbose and matches how the parameters are defined in the papers anyway.

Is there a downside to this I'm missing?  Ie are the more specific class attributes of atoms used downsteam in a different way?


richardjgowers:
Ok cool thanks, I’ll rebase my other PR on top of this

PR #198, opened 2019-01-02
better error messages for missing angles

richardjgowers:
Gives a readout of which angles are missing if you end up with missing angles.  Otherwise kind of frustrating just to be told you're missing x angles... but you guess where.


richardjgowers:
If this isn't problematic I can add a similar thing for dihedrals.  Should be cost-free for 0 errors.  Only problem I can see is maybe truncate number of lines printed (else polymer hell).  For polymer case could instead construct a list of missing types to only print unique missing angle types?

mattwthompson:
I agree it's frustrating, I have a similar setup in #186 where I'm trying to set it up so all the details are printing when some debug option is turned on, but the short (and not so explicitly helpful) messages are printed when it is not.

richardjgowers:
Yeah this is a duplicate

PR #199, opened 2019-01-03
added Forcefield.apply assert_bonds

richardjgowers:


mattwthompson:
Thanks for the contribution!

My understand was that it's supposed to get caught earlier (see [here](https://github.com/mosdef-hub/foyer/pull/155#issuecomment-356496022)) but I know you're dealing with topologically different systems than we were using back then. I'm a little fuzzy on this so I will let @summeraz comment.

richardjgowers:
It looks like that comment says it gets picked up in the MD engine later on.  Arguably this is a thing for foyer to spot.

Looks like this change breaks an existing test, I think you all can push to this branch to fix

richardjgowers:
@summeraz I've added a test, it's possible to do `Forcefield.apply()` and get bonds with no parameters

mikemhenry:
Why did this PR not go through CI?

mattwthompson:
Oops, that was sloppy. We thought it was one of the tests that can fail for silly reasons, but on further inspection it's actually something we need to change. Thanks for catching this

mikemhenry:
No worries! I was just surprised that while I was working on another PR, that test started to fail. It looks like this PR was really needed, since one of the tests didn't have all the bonds defined.

mattwthompson:
Yeah, of all the things that can happen with failing units tests, finding and fixing an unrelated issue is among the better outcomes

Issue #172, opened 2018-06-23
Migrate from plyplus to another parser


mattwthompson:
https://github.com/erezsh/plyplus:

>PlyPlus is now under maintenance only.
> Please check out my new parsing library: https://github.com/erezsh/lark

What this amounts to is the lack of a Python 3.6 release on conda.

I have only touched this portion of the code now, but it looks like it should be a simple switch

https://github.com/mosdef-hub/foyer/blob/b5bf96356c40069f110a907a749be6fb6ed62415/foyer/smarts.py#L80

mattwthompson:
On second thought, it looks like we're hosting the `plyplus` binaries ourselves. I haven't done that before but we could alternatively just add these Python 3.6 binaries.

Similar story in https://github.com/mosdef-hub/mbuild/pull/435

ctk3b:
Would be interesting to try. Rewriting the grammar might be none trivial but we really should put in some work to get to full SMARTS feature parity anyways...

ctk3b:
> it looks like we're hosting the plyplus binaries ourselves. I haven't done that before but we could alternatively just add these Python 3.6 binaries.

This is fairly straight forward and I would recommend this to unblock https://github.com/mosdef-hub/mbuild/pull/435. I believe @justinGilmer and @summeraz have both done it before

justinGilmer:
We have now migrated to `lark-parser`. Its from the same developer as `plyplus`, but is being actively supported. h/t to @volgy for converting this grammar into a `lark` approved syntax.

PR #201, opened 2019-01-05
Support Python 3.6

mattwthompson:
Python 3.6 has been around for a while - and even Python 3.7 isn't too new - so we should start trying to support it.

This will probably require https://github.com/mosdef-hub/mbuild/pull/457 and https://github.com/mosdef-hub/mbuild/pull/435 to be merged first and then an mbuild release.

I am temporarily getting around plyplus issues (#172) by hosting builds on my own Anaconda channel; I will move those to the mosdef channel if it becomes necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ install:
   - conda config --add channels omnia
   - conda config --add channels conda-forge
   - conda config --add channels mosdef
+  - conda config --add channels mwt 
   - conda create -n test-environment python=$PYTHON_VERSION --file requirements.txt
   - source activate test-environment
   - pip install -e .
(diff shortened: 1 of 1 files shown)


mikemhenry:
Anything blocking this PR?

mattwthompson:
Yeah, we need to add each package to Anaconda but we can't do one before doing the other. And host plyplus on the official channel. A shotgun approach of just releasing them separately and hoping their tests pass might work, but that risks putting out builds that don't pass their own tests.

mikemhenry:
Adding the deps needed to a mosdef channel is the right thing to do, you could even added the noarch flag so this doesn't keep happening (or just bump the python version if you don't want to change anything). 

I know hosting these isn't ideal, but many of these seem to be pure python packages that install just fine with pip, is there a way to tell conda to use pip to install them? Right now its really hard to use mBuild and foyer in new projects that I have going that are in python3.6

mattwthompson:
Closed #204

PR #203, opened 2019-02-21
added more python versions to test

mikemhenry:
Start testing and supporting newer python versions


mattwthompson:
This will fail due to (at least) #172. I keep thinking we should schedule a codemonkey session with mulitple of us in the same room/call to try and make more sense of it. At least four of us now have spent time on plyplus/lark individually but collectively gotten nowhere but lost

PR #195, opened 2018-12-11
Run `activate base` before `conda build` in Appveyor

mattwthompson:
Same change as in https://github.com/mosdef-hub/mbuild/pull/477 🤞

PR #204, opened 2019-02-21
Python 3.6 Support

mikemhenry:


mikemhenry:
Okay so I forgot to include an `-y` but this is a good example on how installing foyer + mbuild even with python3.5, it tires to downgrade to 2.7

mikemhenry:
Okay so travis is happy now and things don't get downgraded, I'm a bit grumpy that we will need to have 2 CI scripts to maintain and keep in sync, but until we drop windows I'm sure there is an alternative.

mattwthompson:
I'm gonna mention this PR, since we're doing similar things to what we did in mbuild previously (I actually think we intended to do it for both but simply neglected to do it here)

https://github.com/mosdef-hub/mbuild/pull/457

mikemhenry:
It would be nice to get parity between mbuild and foyer, I think for this I just want to remove the cyclic dependency stuff + not get python downgraded, so then we can do things how its working in mbuild.

mattwthompson:
You're exactly right - the problem is that more needs to be done (see all the changes in the linked PR). I will take it upon myself to make those changes, since we need to release mbuild and we can't do so until similar changes have been made here.

mikemhenry:
Added more python versions to travis so we can see what deps are going to be an issue for python 3.6 and 3.7

.travis.yml
```diff
@@ -24,7 +24,6 @@ install:
   - conda config --add channels omnia
   - conda config --add channels conda-forge
   - conda config --add channels mosdef
-  - conda config --add channels cmelab
```
ctk3b: Is all of omnia's work on conda-forge now?

mattwthompson: I don't think so: https://anaconda.org/search?q=openmm

(OpenMM is our only `omnia` dep, right?)

mikemhenry: AFAIK, yes just openmm, http://docs.openmm.org/latest/userguide/application.html#installing-openmm currently they haven't migrated all over to conda-forge, so you need omnia and conda-forge

mikemhenry:
Still waiting on OSX & windows tests, but it looks like all passes test for python3.6, also added `noarch` to the conda recipe

appveyor.yml
```diff
@@ -7,20 +7,30 @@ environment:
...
   - conda install -yq conda-build jinja2
+  - conda create -n test-environment python=%PYTHON_VERSION% --file requirements-dev.txt
+  - activate test-environment
+  - pip install git+https://github.com/mosdef-hub/mbuild.git
+  - pip install -e .
 
 build: false
 
 test_script:
-  - "%CMD_IN_ENV% activate base"
```
mattwthompson: Any reason why you removed this? I don't recall the specifics but I had to add this at one point to un-break something #195

appveyor.yml
```diff
@@ -7,20 +7,30 @@ environment:
...
     - PYTHON: "C:\\Miniconda36-x64"
       CONDA_PY: "35"
       ARCH: "64"
+    - PYTHON: "C:\\Miniconda36-x64"
+      CONDA_PY: "36"
+      ARCH: "64"
 
 install:
   - set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
+  - conda config --set always_yes yes --set changeps1 no
```
mattwthompson: Is `--set changeps1 no` necessary?

mikemhenry: Nope, its an older artifact I think from where I copied this from, I'll change that

mikemhenry: Actually its to match the travis stuff

mattwthompson:
Since I have the memory of a cinderblock, I'm going to note the things we want to spin off into other small PRs:

* Parity between travis & Appveyor scripts
* Figuring out how appveyor's environments are activated
* Building (but not testing) with noarch

mikemhenry:
Okay things are split out now into separate PRs

mikemhenry:
Use this link check appveyor https://ci.appveyor.com/project/mikemhenry/foyer/builds/22987938

mikemhenry:
3.5 and 3.6 appveyor passed

mattwthompson:
Strange that Python 2.7 failed but I'm not inclined to investigate it much

@justinGilmer we should sit down with Mike to get you up to speed on the many CI issues we've found but the short of this PR is that we can now reasonably expect to ship Python 3.6

mikemhenry:
@mattwthompson This one should be good to go, for the record, we never tested windows python2.7...

Issue #211, opened 2019-03-01
Mark some tests as flaky


mattwthompson:
We often have cron or PR jobs fail due to little things like a network hangup or random weird travis issues (see this from this morning, in which the entire project gets a red badge because a single DOI test failed: https://travis-ci.org/mosdef-hub/foyer/builds/500469282, or the history https://travis-ci.org/mosdef-hub/foyer/builds). Restarting jobs is fine, but it would be nice to avoid this altogether. The badge will probably turn green tomorrow when the next cron jobs run, but it should be green now. I believe marking these tests as flaky is the right approach, @mikemhenry @justinGilmer?

ctk3b:
In general, the test suite should not be making network calls. You should be mocking out the request in the test suite using something like this:
https://requests-mock.readthedocs.io/en/latest/pytest.html

Definitely not necessary here but if you find yourself mocking out lots of requests you can reach for something heavier like this: https://github.com/kevin1024/vcrpy

mikemhenry:
I think we should mock stuff, and then keep in the real network tests but just mark them as flaky, so they can still be tested/catch issues, but won't stop things from passing.

umesh-timalsina:
Despite #276, took into account the changed `crossref` api, tests are failing. Should create mock objects to test when handling the tests. Make sure that the functions using this retry the `request.get`, before giving up.

PR #212, opened 2019-03-03
Mock some requests in tests

mattwthompson:
#211 

The multiple refs test is failing since I can't figure out how to mock it to different values

This is based off of https://realpython.com/testing-third-party-apis-with-mocks, maybe better ways to do this


mattwthompson:
Okay, the first test passes now.

mattwthompson:
This might be something you can do a better job of, @umesh-timalsina? It's a low priority but would be a nice feature.

umesh-timalsina:
Yeah I will take a stab at this PR. Starting with : we should change the way we have written the requests function call assuming that it will always pass [here](https://github.com/mosdef-hub/foyer/blob/90acba0d5ebb9df248eccf58b783d1e29bc6b57a/foyer/forcefield.py#L837).

mattwthompson:
Sorry for the notifications, I am no longer trying to mock stuff but instead using this branch to fix another issue. Hopefully the git history is salvageable enough to anybody trying to solve the original issue of mocking requests.

PR #218, opened 2019-03-11
added no arch so we don't need to make a bunch of seperate builds

mikemhenry:


mattwthompson:
So when we ship to anaconda.org, will this only build one version - does this create a conflict in which it's not clear which versions we support? Because along those lines I would be in favor of building specific versions that we officially support but also a `noarch` version that we throw out to users with no guarantees.

PR #219, opened 2019-03-11
sync up appveyor and travis

mikemhenry:


mattwthompson:
Hmm, it looks like a `$PYTHON_VERSION` variable is getting referenced here but maybe it doesn't exist?

mattwthompson:
I wonder if maybe that's why this was using `conda build`? I don't know how you'd parse `$CONDA_PY` to be in that argument

mattwthompson:
I'm not sure exactly what would be necessary to make this work; while this is a good idea I am not equipped to do the troubleshooting necessary to understand how Appveyor really works

PR #224, opened 2019-03-25
Migrate to lark parser

volgy:
This is the initial version of migrating to the [`lark` parser](https://github.com/lark-parser/lark) form [`plyplus`](https://github.com/erezsh/plyplus) (#172).

### DIFF
--- a/foyer/tests/test_performance.py
+++ b/foyer/tests/test_performance.py
@@ -23,7 +23,7 @@ def test_surface():
 
 
 @pytest.mark.skipif(not has_mbuild, reason="mbuild is not installed")
-@pytest.mark.timeout(60)
+@pytest.mark.timeout(45)
 def test_polymer():
     peg100 = mb.load(get_fn('peg100.mol2'))
     forcefield = Forcefield(name='oplsaa')
--- a/foyer/tests/test_performance.py
+++ b/foyer/tests/test_performance.py
@@ -23,7 +23,7 @@ def test_surface():
 
 
 @pytest.mark.skipif(not has_mbuild, reason="mbuild is not installed")
-@pytest.mark.timeout(45)
+@pytest.mark.timeout(60)
 def test_polymer():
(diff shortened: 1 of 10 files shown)


foyer/tests/test_performance.py
```diff
@@ -23,7 +23,7 @@ def test_surface():
 
 
 @pytest.mark.skipif(not has_mbuild, reason="mbuild is not installed")
-@pytest.mark.timeout(45)
+@pytest.mark.timeout(60)
```
mikemhenry: Is lark slower?

mikemhenry:
LGTM, just have a question about the performance. Installing via pip gave me some issues, but it looks like it will install via conda with this PR.

foyer/tests/test_performance.py
mattwthompson: Yes, I think it's a little slower but it's also more powerful. This specific test was strange because it was consistently failing on only 1 of the 6 tests (macOS Python 2.7) so we decided just to bump it up the tolerance a little; 45s was already somewhat arbitrary and as long as the test runs in a workable amount of time it could be considered passing, IMHO.

mattwthompson:
I don't think a user would normally try and install this dep, but if there is reason to expect one to, we should note that `conda install -c conda-forge lark-parser` is the way to go. But the CI seems to have no issue with it, which is a good sign.

mattwthompson:
@mikemhenry I think you told me elsewhere but could you confirm that this works locally for one of the messy systems Eric's lab deals with?

foyer/tests/test_performance.py
volgy: I switched to the LALR parser in b06d372f2737dd7af5a1046fee3b9636d036970e. Based on my tests this results in ~3x speed improvements (with the current test suite)

mattwthompson:
Any @mosdef-hub/mosdef-contributors want to take a look at this? Will be merging in soon if nothing troubling pops up

Issue #205, opened 2019-02-21
Dependency Triage 


mikemhenry:
In efforts to make foyer easier to install, we should audit what dependencies we are pulling in. Additionally, keeping dependencies light keeps us from being stuck when upstream doesn't want to support a newer python version. 

For these packages, we need to

1. Summarize what they do for us
1. Evaluate how supported they are/updated
1. Evaluate if we can implement this ourselves

- [ ] parmed
- [ ] networkx
- [ ] oset
 - [ ] six
 - [ ] openmm <7.3
 - [ ] plyplus
 - [ ] requests
 - [ ] lxml

Then we can talk about what we want to cut/implement ourselves moving forwards. This list also is not authoritative, but taken from https://github.com/mosdef-hub/foyer/blob/master/devtools/conda-recipe/meta.yaml so I would appreciate it if someone double check to make sure I didn't miss anything (or if something can be removed since its not used at all)

ctk3b:
Drive by comments:

* `oset`
Everything you can do with `oset`, you can do with the built-in `OrderedDict`. It's just not quite as clean.

Alternatively, boltons has an `IndexedSet`, is better maintained and is on conda-forge: https://anaconda.org/conda-forge/boltons

* `requests`

This package is about as ubiquitous as it gets in the python community. Unless you want to remove the feature to pull forcefields via a URL, I think this should be easy enough to keep.

* `six`

Just drop python 2 support ;)

* `networkx`

Don't think there's a good way around this.

* `plyplus`

Might be possible to swap it out for something else but you do NOT want to build this yourself.

* `lxml` 

Might be possible to use the built-in xml module. Maybe check what openmm uses since that's the format we're working with.

mattwthompson:
Of those, we've only encountered issues with `plyplus` - maybe others have also died but they see fine, at least for now.

Stripping the dependencies into "normal" and "dev" helps a lot - the basic requirements list isn't too bad (and will be one shorter when we formally stop caring about Python 2.7 😄)

summeraz:
Echoing @ctk3b's comments. These are all pretty standard packages, minus `parmed` and `openmm` which we obviously need to keep, and `plyplus`. @mattwthompson outlined our need to migrate from `plyplus` in #172, and I think we should be able to just replace `plyplus` with `lark` which has a noarch package on conda-forge. I'll try to take a whack at implementing that change.

mattwthompson:
> I'll try to take a whack at implementing that change

[...]

PR #226, opened 2019-03-26
Switch from oset OrderedSet to boltons IndexedSet

mattwthompson:
In the spirit of using packages that have been updated in the past 5 years (https://pypi.org/project/oset/#history)

#205

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -18,7 +18,7 @@ requirements:
     - python
     - parmed
     - networkx >=2.0
-    - oset
+    - boltons
     - six
     - openmm <7.3
     - plyplus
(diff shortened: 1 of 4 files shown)


ctk3b:
Quick question: do we still rely on the ordered nature of the white/black list? A quick scan through the code suggests no. Perhaps it _slightly_ aids when debugging?

mattwthompson:
You're more familiar with that than me but I agree with your assessment that the order doesn't seem to matter at the moment. Were those lists `OrderedSet` instead of `set` before because it used a different way of solving the atomtyping problem?

mattwthompson:
Just using built-in `set` works for me locally, I will open that as a PR if the build passes on my fork https://travis-ci.org/mattwthompson/foyer/builds/518369398

ctk3b:
> Were those lists OrderedSet instead of set before because it used a different way of solving the atomtyping problem?

I honestly don't remember. At the end of the day, we always used a set diff to check if we had converged. I suspect we had used this to aid in debugging back when this was a lot less stable and we were still working on edge cases in the atomtyping logic itself.

justinGilmer:
Taken care of by #231

PR #229, opened 2019-03-29
Backend integration

ahy3nz:
Don't expect any of the tests to pass, but opening this now for visibility, contributions, comments, and critiques. 

Overarching changes
* The goal is to move away from any dependence on `OpenMM` and `ParmEd`
* This means parsing the XML files ourselves (I have a primitive XML parser in `forcefield.py`, but this will depend on how/if we plan to expand XML functionality). This parser will generate a bunch of `topology.Potential` objects, stored in `foyer.Forcefield` object
* `foyer.Forcefield` no longer inherits from `omm.Forcefield`. I took a first pass at avoiding/reworking any calls that would have depended on the `omm.Forcefield` superclass
* Atomtyping functionality was largely unchanged, I took a first pass at some refactoring to iterate through `topology.Topology` and not `openMM.Topology`

My proposed implementations (discussion greatly appreciated here):
* We're parsing an XML file to create a bunch of `topology.Potential` objects into a variety of attributes in `foyer.Forcefield`
* After atomtyping, every `topology.Site` should have an associated atomtype (this is NOT `topology.AtomType`, but rather the string that I __think__ gets stored as `atom.id` (now `site.id`)
* Most of the work comes from actually parametrizing the `topology.AtomTypes, BondTypes, AnglesTypes`
* Stitch together the appropriate `foyer.Forcefield.atomtypes` (list of `topology.AtomType`) with `topology.Site.atom_type` (parametrizing the atomtype)
* Assuming all bonds, angles, etc are enumerated in `topology.Topology`, the constituent `topology.Sites` should all have `topology.AtomTypes` associated, so now we need to stitch together the appropriate `foyer.Forcefield.bondtypes` (list of `topology.BondType`) with `topology.Bond.connection_type` (parametrizing the bondtype)

The general XML schema i was parsing was hoping to allow for specifying different BondType attributes at different levels within the etree. I'm currently not sure the best way to handle unyt.units, but there's a function in `foyer.Forcefield` that tries to parse the unyt.units from these xml attributes
```
<ForceField>
    <BondType expression="0.5 * k * (r - r_eq)" independent_variables="r">
        <BondType k="50 * kcal/(nm**2)" r_eq="200*nm"/>
        <BondType k="100 * kcal/(nm**2)" r_eq="20*nm"/>
        <BondType k="150 * kcal/(nm**2)" r_eq="2*nm"/>
        <BondType k="200 * kcal/(nm**2)" r_eq="0.2*nm"/>
    </BondType>
    <BondType expression="0.5 * k * (r - r_eq)" k="1 * kcal" r_eq="1 * nm"
        independent_variables="r"/>
    <BondType expression="0.5 * k * (r - r_eq)" k="2 * kcal" r_eq="2 * nm"
        independent_variables="r"/>
    <BondType expression="0.5 * k * (r - r_eq)" k="3 * kcal" r_eq="3 * nm"
        independent_variables="r"/>
</ForceField>
```

### DIFF
--- a/foyer/atomtyper.py
+++ b/foyer/atomtyper.py
@@ -97,7 +97,7 @@ def _resolve_atomtypes(top, forcefield):
         if len(atomtype) == 1:
             #atom.id = atomtype[0] # this is just a string
             atomtype_param = [atype for atype in forcefield.atomtypes 
-                    if atype.name==atomtype]
+                    if atype.name==atomtype[0]]
             atom.atom_type = atomtype_param[0]
         elif len(atomtype) > 1:
             raise FoyerError("Found multiple types for atom {} ({}): {}.".format(
--- a/foyer/atomtyper.py
+++ b/foyer/atomtyper.py
@@ -4,12 +4,12 @@ from foyer.exceptions import FoyerError
 from foyer.smarts_graph import SMARTSGraph
 
 
-def find_atomtypes(topology, forcefield, max_iter=10):
+def find_atomtypes(top, forcefield, max_iter=10):
     """Determine atomtypes for all atoms.
 
     Parameters
     ----------
-    topology : topology.Topology
+    top : topology.Topology
         The topology that we are trying to atomtype.
     forcefield : foyer.Forcefield
         The forcefield object.
@@ -21,7 +21,7 @@ def find_atomtypes(topology, forcefield, max_iter=10):
 
     # Only consider rules for elements found in topology
     subrules = dict()
-    system_elements = {a.name for a in topology.sites} # We don't have any elements in topology, so I think this might be okay?
+    system_elements = {a.name for a in top.sites} # We don't have any elements in topology, so I think this might be okay?
     for key,val in rules.items():
         atom = val.node[0]['atom']
         if len(atom.select('atom_symbol')) == 1 and not atom.select('not_expression'):
@@ -39,8 +39,8 @@ def find_atomtypes(topology, forcefield, max_iter=10):
             subrules[key] = val
     rules = subrules
 
-    _iterate_rules(rules, topology, max_iter=max_iter)
-    _resolve_atomtypes(topology)
+    _iterate_rules(rules, top, max_iter=max_iter)
+    _resolve_atomtypes(top)
 
 
 def _load_rules(forcefield):
@@ -59,7 +59,7 @@ def _load_rules(forcefield):
     return rules
 
 
-def _iterate_rules(rules, topology, max_iter):
+def _iterate_rules(rules, top, max_iter):
     """Iteratively run all the rules until the white- and backlists converge.
 
     Parameters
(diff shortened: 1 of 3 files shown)


mattwthompson:
Super exited about this! My main worry here is that this will become a behemoth of a PR, so I think we should try to split this up into smaller pieces. It's hard to think about exactly how to do this - maybe starting with just the XML parser?

> We're parsing an XML file to create a bunch of topology.Potential objects into a variety of attributes in foyer.Forcefield

This seems like the best starting point

> After atomtyping, every topology.Site should have an associated atomtype (this is NOT topology.AtomType, but rather the string that I think gets stored as atom.id (now site.id)

Does this attribute refer to i.e. `opls_142`? Because we'd want to create a different name to be clear that it's just a string representation that should point to an actual `Potential` object elsewhere (or maybe wrap it into a `Potential` object and connect them through this attribute?)

> Most of the work comes from actually parametrizing the topology.AtomTypes, BondTypes, AnglesTypes
Stitch together the appropriate foyer.Forcefield.atomtypes (list of topology.AtomType) with topology.Site.atom_type (parametrizing the atomtype)

For sure, this is where performance comes into play. My experience with the current backend is that it's reasonably fast but could be improved on. I don't know how much of a focus performance was in developing `ParmEd` - it's clearly good enough for any reasonably-sized system so I don't know if doing multi-million-atom systems was a focus.

> I'm currently not sure the best way to handle unyt.units

I think the easiest way to get this started would be to specify units at some higher level and not use them in the lowest levels of the tree, i.e. this, if it can be done (try to ignore my bad syntax)


```
<ForceField, distance_unit='nm', energy_unit='kcal'/>
  <BondTypes expression="0.5 * k * (r - r_eq)" independent_variables="r"/>
    <Bond type1="foo" type2="foo" k="50" r_eq="0.15/>
    <Bond type1="foo" type2="foo" k="200" r_eq="0.15/>
</ForceField>

ahy3nz:
I was able to write a toy FFXML (atomtypes, bondtypes, angletypes), create a `topology.Topology` object, add sites and bonds to create a methane, use the new XML parsing to create the new `foyer.Forcefield` object, atomtype through foyer, and apply bond/angle parameters through foyer. The resultant, parametrized `topology.Topology` object has all atomtypes,bondtypes, and angletypes, with a references file (but this requires the pending PR hashing functionality in `topology` objects). I don't think there are any calls to OpenMM generator/forcefield/simulation objects

Things this PR might still need:
* Residue mappings/concept of residues aren't available in `topology.Topology` yet, so that' been commented out.
* There's probably more refactoring that will be needed for more sophisticated smarts functionality (the toy methane example is definitely not a good stress test of smarts functionality)
* Expanding the supported potential/FF terms (currently only looking for BondTypes and AngleTypes, nothing about DihdralTypes)
* The XML style still needs discussion on an agreed-upon syntax, including units specification
* #224 migrating to lark will change a bunch of methods too, hopefully without too many conflicts

PR #231, opened 2019-04-10
Use built-in sets for whitelists & blacklists

mattwthompson:
See #226 for discussion and #205 for context. `boltons` seems to be a light dep but there doesn't seem to be a compelling reason to need order/indexing/etc in the whitelist blacklist. Maybe it could help in debugging but that part of the package seems pretty stable now.

Probably squash & merge this one, or I can clean up the git history.


justinGilmer:
This closes #226

PR #235, opened 2019-04-27
Support NBFixes

mattwthompson:
#128 

This starts the implementation I had in mind from our discussions yesterday @ahy3nz. I haven't yet figured out the best way to access the individual `NBFixPair` section of the XML in memory because it requires figuring out exactly how OpenMM does it.

Here is a messy function taken from another repo that demonstrates how I (will) add these into the output:

```
def _apply_nbfixes(struct):
    """"Add the nbfixes into the structure."""
    nbfixes = parse_nonbond_txt(get_fn('cross_interactions.txt')) # Here we will grab them from FF.something
    atom_types = list({a.atom_type for a in struct.atoms})
    for atom_type in sorted(atom_types, key=lambda a: a.name):
        for nbfix_pair in [*nbfixes.keys()]:
            if atom_type.name in nbfix_pair:
                if atom_type.name == nbfix_pair[0]:
                    a2 = nbfix_pair[1]
                if atom_type.name == nbfix_pair[1]:
                    a2 = nbfix_pair[0]
                sig, eps = nbfixes[nbfix_pair]
                sig = sig.in_units_of(angstrom)._value
                rmin = sig * 2 ** (1 / 6)
                eps = eps.in_units_of(kilocalorie)._value
                atom_type.add_nbfix(a2, rmin, eps)
```


mattwthompson:
Closing in favor of #267 short-term and the new backend long-term

Issue #222, opened 2019-03-12
Better validation of XML files


mattwthompson:
We should try to avoid bugs like #221 by checking that type's masses are similar to the atomic mass of the element. This would be limited to atomistic systems.

I'm also adding the `validation` tag because it seems like we had a lot of discussions related to validating various steps of the process - pre-processing XML files, validating the atomtyping process, checking the results

PR #236, opened 2019-05-02
Carbon with incorrect mass

### LINKED ISSUE #221: wrong mass on line 758 for oplsaa

justinGilmer:
### PR Summary:
@mikemhenry pointed out in #221 that a carbon atom had an incorrect mass.

This has been fixed.

### PR Checklist
------------

closes #221


mikemhenry:
This is a good fix, but how about we add a unit test that checks all the masses? I'm not an oplsaa expert, but should all the masses correspond to their atomic mass?

mattwthompson:
I don't think anybody is truly an OPLS expert but I agree this would be a nice feature. That's the idea behind #222, but I'm not sure the best way to structure it, since it would require grabbing that data to compare against.

mikemhenry:
Got it!

Issue #239, opened 2019-05-20
Update PeriodicTorsionForce writer for xml_writer


ahy3nz:
**Bug summary**
#161 only looks for the first term in periodic torsion forces (`periodicity1` and not `periodicity2` etc). Sometimes additional periodicities in the torsions exist, but sometimes they don't. The resultant XML file should print out all the associated periodicities associated with a particular torsion.

What were you trying to do and what happened instead? Please copy and paste the stack output
Since OpenMM "unwraps" each periodicity into its own torsion force, each periodicity gets separated, and the resultant `pmd.Structure` will generate a separate `pmd.Dihedral` for each periodicity. This also breaks the uniqueness-identification in printing dihedral-XML lines - what SHOULD be a single XML periodictorsionforce line for a single 4-bonded-tuple with multiple periodicities becomes multiple XML lines; when looking for uniqueness, each periodicity will have different parameters and we end up with a lot of redundant dihedral XML lines

**Code to reproduce the behavior**

Please include a code snippet that can be used to reproduce this bug.

An example "original" ffxml could have 2 periodicities defined for a given PeriodicTorsion
```
<Proper type1="opls_140" type2="opls_135" type3="opls_135" type4="opls_140" per    iodicity1="1" phase1="3.14" k1="3.1" periodicity2="2" phase2="1.57" k2="6.7"/>
```

But the "written" ffxml will unwrap all the periodic terms, and do so redundantly
```python
<PeriodicTorsionForce>
      <Proper type1="opls_140" type2="opls_135" type3="opls_135" type4="opls_140" per    iodicity1="1" phase1="3.14" k1="3.1"/>
      <Proper type1="opls_140" type2="opls_135" type3="opls_135" type4="opls_140" per    iodicity1="2" phase1="1.57" k1="6.7"/>
      <Proper type1="opls_140" type2="opls_135" type3="opls_135" type4="opls_140" per    iodicity1="1" phase1="3.14" k1="3.1"/>
      <Proper type1="opls_140" type2="opls_135" type3="opls_135" type4="opls_140" per    iodicity1="2" phase1="1.57" k1="6.7"/>
```
Notice these lines all refer to the same 4-bonded tuple, but with lots of redundancy

**Software versions**

- Which version of Foyer are you using? (`python -c "import foyer; print(foyer.version)"`) 0.5.2
- Which version of Python (`python --version`)? 3.6.7
- Which operating system? Mac

[...]

PR #241, opened 2019-05-21
Periodictorsions xml

ahy3nz:
### PR Summary:
#239 This code modifies the XML writer to look for multiple periodictorsionforces that apply to the same dihedrals. This is done by comparing the "current" periodic dihedral to the "previous" periodic dihedral, and determining whether the two dihedral forces should be merged into one. 

As a side note, the issues with the DOI XMLprinting should have been appropriately addressed in #161, any issues we saw with the DOI XMLprinting were due to flaws in the "source" FFXML we were using to parametrize these structures, not in how we were printing to the "sink" FFXML

### PR Checklist
------------

### DIFF
--- a/foyer/tests/test_forcefield.py
+++ b/foyer/tests/test_forcefield.py
@@ -295,3 +295,10 @@ def test_write_xml_multiple_periodictorsions(filename):
     typed_struc = ff.apply(cmpd, assert_dihedral_params=False)
     typed_struc.write_foyer(filename='multi-periodictorsions.xml', forcefield=ff, unique=True)
 
+    partial_ff = Forcefield(forcefield_files='multi-periodictorsions.xml')
+    typed_by_partial = partial_ff.apply(cmpd, assert_dihedral_params=False)
+
+    assert len(typed_struc.bonds) == len(typed_by_partial.bonds)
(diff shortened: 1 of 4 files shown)


ahy3nz:
Build is passing, looking for comments (and if the resultant XMLs look right to other eyes)

foyer/tests/test_forcefield.py
```diff
@@ -287,3 +287,18 @@ def test_write_xml(filename):
...
+    typed_struc = ff.apply(cmpd, assert_dihedral_params=False)
+    typed_struc.write_foyer(filename='multi-periodictorsions.xml', forcefield=ff, unique=True)
+
+    partial_ff = Forcefield(forcefield_files='multi-periodictorsions.xml')
+    typed_by_partial = partial_ff.apply(cmpd, assert_dihedral_params=False)
+
+    assert len(typed_struc.bonds) == len(typed_by_partial.bonds)
+    assert len(typed_struc.angles) == len(typed_by_partial.angles)
+    assert len(typed_struc.dihedrals) == len(typed_by_partial.dihedrals)
+
```
mattwthompson: Can we also so some checking to make sure that the variables like `periodicity2`/`phase2`/`k2` variables still exist in the force field object after it's read back in (and ideally are of similar value as well)?

PR #242, opened 2019-05-22
Move tests into classes so they can inherit pytest fixtures

mattwthompson:
### PR Summary:
We now have a few unit tests that check to see files can be written out. If you've seen a few `.bib` and XML files floating around the root directory of your `foyer` install, it's because these tests did not include the `pytest` fixture that chucks everything into a temporary directory and nukes it after the test completes. This PR makes that change, which isn't necessary for most tests but is a good practice in case other fixtures are added in the future.

Maybe 90% of these changes are tabbing unit tests over 4 spaces; I'm not sure why GitHub doesn't make the diffs look prettier than they do.

### PR Checklist
------------

### DIFF
--- a/foyer/tests/base_test.py
+++ b/foyer/tests/base_test.py
@@ -1,0 +1,8 @@
+import pytest
+
+
+class BaseTest:
+
+    @pytest.fixture(autouse=True)
+    def initdir(self, tmpdir):
+        tmpdir.chdir()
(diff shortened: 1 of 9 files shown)


ctk3b:
I think you can achieve the same by just marking the fixture as `@pytest.fixture(scope="session")`

mattwthompson:
I think the counterpoint is that having things in classes brings it in to parity with how mbuild tests are structured and makes it easier to use other pytest plugins in the future

mattwthompson:
Maybe classes will be useful in the future but for now a `conftest.py` file works #245

PR #246, opened 2019-05-26
Openmm 73 patch

ahy3nz:
### PR Summary:
Addresses #188 . This allows the use of Openmm 7.3 in with Foyer. The issue was the `createSystem`. OMM 7.2 didn't have default kwargs for `switchDistance`; in OMM 7.3, their `omm.FF.createSystem()` now added a default kwarg for `switchDistance`, which got added to the arguments passed to `force.createForce()`. I think we just needed to add a default kwarg to `switchDistance` in `foyer.FF.createSystem()` such that this argument gets passed into `force.createForce()` in the Foyer code. 

There are some extra commits and comments in the code here - I have multiple "options" to accommodate this switchDistance issue. If we go with the default kwarg in `foyer.ff.createSystem()`, we don't need to modify any tests. Some other options/code I explored were manually specifying the `switchDistance=None` in some parts of `apply()` routine, but that would then involve modifying `test_residue_map` because that unit test doesn't actually use the entire `apply()` routine, but rather snippets of the code inside the `apply()` routine. 

Tests passed locally on a py37 environment with `openmm=7.3.1` I don't think I need to add a unit test?
### PR Checklist
------------

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,11 @@ matrix:
     - { os: linux, env: PYTHON_VERSION=2.7 }
     - { os: linux, env: PYTHON_VERSION=3.5 }
     - { os: linux, env: PYTHON_VERSION=3.6 }
+    - { os: linux, env: PYTHON_VERSION=3.7 }
     - { os: osx, env: PYTHON_VERSION=2.7 }
     - { os: osx, env: PYTHON_VERSION=3.5 }
     - { os: osx, env: PYTHON_VERSION=3.6 }
+    - { os: osx, env: PYTHON_VERSION=3.7 }
 
 env:
   global:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,6 +13,10 @@ environment:
     - PYTHON: "C:\\Miniconda36-x64"
       CONDA_PY: "36"
       ARCH: "64"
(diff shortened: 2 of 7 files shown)


mattwthompson:
Thanks!

Since the concept of `switchDistance` doesn't exist in the output of `.apply()` I'm not worried about the details as long as it produces a sensible output.

mattwthompson:
Also works for me in a local environment (Python 3.7 and Openmm 7.3.1)

ahy3nz:
> Since the concept of `switchDistance` doesn't exist in the output of `.apply()` I'm not worried about the details as long as it produces a sensible output.

As far as I know (someone correct me if I'm wrong), even though now we're explicitly accounting for `switchDistance=None`, I don't think this is going to change any of the various conversions between mbuild/foyer/openmm/parmed, nor will it affect any of the underlying molecular mechanics information. 

It could be nice to see if the simulation workflows before and after this PR remain unchanged.

As far as new commits go, I think I'm done altering the foyer API (unless a bug is found in this PR). Any new commits will be me attempting to change the devtools such that we test python 3.7 and unpin the OpenMM version

ahy3nz:
Merged in  #247

PR #252, opened 2019-06-06
Enforce deletion of temporary XML files

### LINKED ISSUE #244: Temp files when pre-processing FF xml files
**Bug summary**

in `forcefield.py` we pre-process all the FFXMLs, THEN load them into the `foyer.Forcefield` object. In pre-processing, we create a bunch of tempfiles (`mkstemp`) in a tempdir, then load all the tempfiles.

[after this line, where we create the super from the preprocessed files](https://github.com/mosdef-hub/foyer/blob/f140229492555ebb35d56b53139e8c0c8f82df4f/foyer/forcefield.py#L305) probably needs a bunch of `os.unlink` calls on the pre-processed files, like in `mbuild.packing.py`. So we still need to make a tempdir with temp files (maybe we can work around this, but for another PR), load them into the FF object, then we need to delete them

justinGilmer:
### PR Summary:
Change the temporary file creation from `tempfile.mkstemp` to
`tempfile.NamedTemporaryFile`.

`mkstemp` creates a temporary file with a file handle that can be
accessed in a secure manner, preventing possible race conditions.
However, `mkstemp` also requires _manual_ deletion of the file
after its purpose has been served.

This file was never being deleted, and would pollute the `$TMPDIR`
of the user who was running `Foyer`.

Instances of `mkstemp` have been replaced with `NamedTemporaryFile`,
which provides a higher level interface to created temp files,
and all instances where these are used are captured in `try,finally`
blocks, to ensure that they are removed after the XML files have been
processed.

Resolves #244 

### PR Checklist
------------

### DIFF
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -2,7 +2,7 @@ import collections
 import glob
 import itertools
 import os
-from tempfile import mktemp, mkstemp
+from tempfile import NamedTemporaryFile
 import xml.etree.ElementTree as ET
 
 try:
@@ -76,12 +76,12 @@ def preprocess_forcefield_files(forcefield_files=None):
                           'objects by precedence.')
 
         # write to temp file
-        _, temp_file_name = mkstemp(suffix=suffix)
-        with open(temp_file_name, 'w') as temp_f:
+        temp_file = NamedTemporaryFile(suffix=suffix, delete=False)
+        with open(temp_file.name, 'w') as temp_f:
             temp_f.write(xml_contents)
 
         # append temp file name to list
-        preprocessed_files.append(temp_file_name)
+        preprocessed_files.append(temp_file.name)
 
     return preprocessed_files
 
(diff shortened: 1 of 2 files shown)


justinGilmer:
I tested this locally.

To test:
* Run the unit tests for `foyer` without this change
* check for leftover tmp xml files
- `ls $TMPDIR/*.xml`
* delete the files
- `rm $TMPDIR/*.xml`
Apply this change
* run the tests again
* check for leftover tmp xml files
- `ls $TMPDIR/*.xml`

This should return an error, as there are no xml files to find

justinGilmer:
Just for an  extreme example, I was using a local machine to atomtype ~30K systems, my tmp folder afterwards was filled with approximately ~7GB of tmp xml files.

mattwthompson:
Works for me locally.

In:

```
git checkout master                                                                                                                                    
rm $TMPDIR/*xml    
py.test -q --disable-warnings    
ls $TMPDIR/*xml | wc -l    
rm $TMPDIR/*xml    
git checkout file_open_fixes    
pip install -e .    
py.test -q --disable-warnings    
ls $TMPDIR/*xml | wc -l 
```

Out:
```
(master)$ bash test_file_open_fixes.sh 
Already on 'master'
Your branch is up to date with 'upstream/master'.
............................................................................................................................................... [ 93%]
..........                                                                                                                                      [100%]
153 passed, 172 warnings in 64.49 seconds
     182
Switched to branch 'file_open_fixes'
Obtaining file:///Users/mwt/software/foyer
Installing collected packages: foyer
  Found existing installation: foyer 0.6.2
    Uninstalling foyer-0.6.2:
      Successfully uninstalled foyer-0.6.2
  Running setup.py develop for foyer
Successfully installed foyer
............................................................................................................................................... [ 93%]
..........                                                                                                                                      [100%]
153 passed, 172 warnings in 50.33 seconds
ls: /var/folders/2w/dfnzsrm57_730c_n_h0f3psc0000gn/T//*xml: No such file or directory
       0

Issue #240, opened 2019-05-21
Add overrides to xml_writer


ahy3nz:
**Describe the behavior you would like added to Foyer**
The current xml_writer does not print out the overrides. This is an important feature to include. If a system has aliphatic alkenes and aromatic alkenes, the aromatic alkene may satisfy both the aliphatic alkane and aromatic alkene atomtype definitions - overrides would be necessary to resolve the atom type

**Describe the solution you'd like**
Similar logic as [here](https://github.com/mosdef-hub/foyer/blob/4787e68d45466e8aec19a4f99db07c65f1ebfd8e/foyer/xml_writer.py#L81)
A very simple line like `('overrides', 'forcefield.atomTypeOverrides[name]')` might be all that is necessary, and then some string-formatting to look nice in the XML

**Describe alternatives you've considered**
For some systems in which there aren't likely to be clashes, this probably won't be a problem. But it seems important to be able to use whitelists/blacklists/overrides

**Additional context**

uppittu11:
Looks like this was fixed

PR #254, opened 2019-06-11
Update xml writer

daico007:
### PR Summary:
Add overrides feature for xml_writer as suggested in #240. Note that in order to avoid ValidationError, we only gonna write `overrides = "{'atom_type_n1','atom_type_n2'}"` if `atom_type_n1` and `atom_type_n2` are in atoms' `atom_type`. 

### PR Checklist
------------


foyer/xml_writer.py
```diff
@@ -95,6 +99,18 @@ def _write_atoms(self, root, atoms, forcefield, unique):
                 if key == 'doi':
                     label = eval(val)#[a for a in label]
                     label = ','.join([a for a in label])
+                elif key == 'overrides':
+                # Only write overrides atomtypes if they are in self.atom_type
+                    label = []
+                    for item in eval(val):
```
ctk3b: This is largely unrelated to this PR but what _necessitated_ the use of `eval` here?

What is the shape of data that would raise `AttributeErrors` and `KeyErrors`?

daico007: To be honest, I am not really sure. Here I am just kind of follow the earlier convention in this block of code. 
I will try strip off the eval part and see if it makes any difference.

daico007: So here is what happened when I use `for item in val:`

```<Type name="opls_145" class="CA" element="C" mass="12.01078" def="[C;X3;r6]1[C;X3;r6][C;X3;r6][C;X3;r6][C;X3;r6][C;X3;r6]1" overrides="f,o,r,c,e,f,i,e,l,d,.,a,t,o,m,T,y,p,e,O,v,e,r,r,i,d,e,s,[,n,a,m,e,]" desc="" doi="10.1021/ja9621760"/>```

From the block above where we define `atomtype_attrs = collections.OrderedDict([key,val])`, `val` here is actually `forcefield.atomTypeOverrides[name]`, and hence we need the `eval` step to return a primitive data type.

**Comment by daico007** [comment]
If everything is evaluated to `True`: 
```
<AtomTypes>
    <Type name="opls_145" class="CA" element="C" mass="12.01078" def="[C;X3;r6]1[C;X3;r6][C;X3;r6][C;X3;r6][C;X3;r6][C;X3;r6]1" overrides="opls_142,opls_141" desc="" doi="10.1021/ja9621760"/>
    <Type name="opls_146" class="HA" element="H" mass="1.007947" def="[H][C;%opls_145]" overrides="opls_144" desc="" doi="10.1021/ja9621760"/>
</AtomTypes>
```

**Comment by daico007** [comment]
If the `if` evaluated to be `False` (so `overrides` 's `atom_type` is not in atoms' `atom_type`)

```
<AtomTypes>
    <Type name="opls_145" class="CA" element="C" mass="12.01078" def="[C;X3;r6]1[C;X3;r6][C;X3;r6][C;X3;r6][C;X3;r6][C;X3;r6]1" overrides="" desc="" doi="10.1021/ja9621760"/>
    <Type name="opls_146" class="HA" element="H" mass="1.007947" def="[H][C;%opls_145]" overrides="" desc="" doi="10.1021/ja9621760"/>
</AtomTypes>
```

File: foyer/xml_writer.py (hunk 1 above)
Comment (ctk3b) [reply foyer/xml_writer.py]: Ah I see. No need to hold up your current PR but a pattern for avoiding this would be to define `getter` functions for every attribute.

They would look roughly like `get_XYZ(atom)`. The override function for example could look like:

```python
def get_atomtype_overrides(atom):
    return forcefield.atomTypeOverrides[atom.name]
```

The loop would be rewritten as:
```python
for attribute, getter in atomtype_attribute_getters.items():
   if attribute == 'overrides':
        for item in getter(atom):
           ...
```

ahy3nz:
Can you put together a unit test to test that we can correctly print out the overrides? Take a system, atomtype it (hopefully the atomtypes will require the use of overrides), print out the xml, validate that the "overrides" xml-attribute is indeed getting written in, and see that it the resultant xml is sufficient to atomtype another system?

daico007:
@ahy3nz Sure! I believe this one currently is partially covered by `test_write_xml` function in `test_forcefield.py` (At least, I bumped into the `ValidationError: Reference to undefined atomtype` several times for the `overrides` of benzene)
I will try to find a few other cases where the `overrides` is actually got written out and add them to the test function.

daico007:
@ahy3nz Update, I still cannot find a `.mol2` file that satisfies the `if item in atom_type_dict:` statement to naturally print out the `overrides=""`. I understand that this is due to blacklist/whitelist mechanism, but I can't find the special case as mentioned in #240, so I can't really add a test for this `overrides` feature.

Also, I talked with @justinGilmer on Friday about whether it is desirable to write out the original overrides atom_type as a comment in the output file, which would look like: 
```
<Type name="opls_145" class="CA" element="C" mass="12.01078" def="[C;X3;r6]1[C;X3;r6][C;X3;r6][C;X3;r6][C;X3;r6][C;X3;r6]1" desc="" doi="10.1021/ja9621760" overrides="">
      <!--Note: original overrides="opls_142,opls_141"-->
</Type>
```
So I really want to know what everybody thinks about this, do you think this information is helpful (as for somebody who wants to replicate a project,etc), or is it just reduntdant?

ahy3nz:
Okay just summarizing some discussion: 

In these XMLs, an atomtype listed in an `overrides` section must have a corresponding XML element defining that atomtype. Similar to #255 where the extraneous extraneous atomtypes are trimmed from the smarts definition, these extraneous atomtypes get trimmed from the list of atomtypes. Personally, I like that these extraneous atomtype overrides get written as comments in the XML.

And maybe some sort of unit tests that verifies some atomtypes get written to the overrides or this XML comment

Issue #255, opened 2019-06-11
Undefined atom types referenced in SMARTS definition from xml writer


rmatsum836:
**Bug summary**
Testing #254, I tried reading in the xml file that was output from `write_foyer`.  
When attempting to create a `Forcefield` object from this xml, an error occurs due to undefined atom types being referenced in the SMARTS definitions.

Based on this error, we should parse out the atom types that don't exist in the xml from the SMARTS definitions. 

**Code to reproduce the behavior**

```python
lopes = Forcefield('co-test.xml')
foyer.exceptions.ValidationError: Reference to undefined atomtype 'lopes_016' in SMARTS string 'H[C;%lopes_016,%lopes_017,%lopes_018,%lopes_019]' at line 12
```
The issue is that `lopes_016` isn't an atom type in this specific system, so it isn't written out to the xml from `write_foyer`.  

**Software versions**

- Which version of Foyer are you using? 0.6.2
- Which version of Python (`python --version`)? 3.7.3
- Which operating system? Mac OS 10.14.4

daico007:
@rmatsum836 Can you upload the `co-test.xml` file?

rmatsum836:
https://gist.github.com/rmatsum836/568a1c7218f127542ac8e8fd3c6b485a

ahy3nz:
I think there are 3 options:
1) Turn off the `Validator` just for the sake of testing this

2) If an atomtype is referenced in a smarts description, then include that atomtype in the ffxml. 

2a) Include the extra atomtype, but not the smarts string. This extra atomtype might not need a sophisticated smarts string if this atomtype is never even used in a particular chemical system. Really this option would just get around any complaints from the `Validator`. While the extra atomtype would be present, it might never get used?

2b) Include the extra atomtype, but include the smarts string. This would be a more "complete" ffxml, but then you'd have to add even more atomtypes depending on the extra atomtype's smarts string

3) Modify the smarts string to trim out any atomtypes that aren't present in the chemical system? Not sure if this will incur any consequences if we modify the smarts definitions so the minimal-FFXML is purely self-described

rmatsum836:
Option 3: modifying the smarts string to to trim out any non-existent atom types was the first approach I tried.  It's not immediately clear to me how to parse the string in a general enough manner, so I might go with 2b.

daico007:
[...]

PR #256, opened 2019-06-12
Remove undefined atom types from SMARTS strings in xml writer.

rmatsum836:
### PR Summary:
This is to address #255 where atom types were being referenced in the SMARTS strings without being defined.  The approach I took was to build up a list of undefined atom types that are referenced in SMARTS strings and then strip them away from the SMARTS strings.  

I believe that this only occurs in SMARTS strings where the `or` logic is applied with a comma.

In addition, I added some basic unit tests for the xml writer functionality.  I need to find an example of this edge case where I can atomtype a Parmed structure, write it out to an xml file, and then load it back in.  I've tested this PR with an unpublished force field, and it seems to work thus far.
### PR Checklist
------------


foyer/xml_writer.py
```diff
@@ -108,6 +110,37 @@ def _write_atoms(self, root, atoms, forcefield, unique):
...
+                                   name=name)
+        for atom_expr in nx.get_node_attributes(smarts_graph, name='atom').values():
+            labels = atom_expr.find_data('has_label')
+            for label in labels:
+                atom_type = label.children[0][1:]
+                smarts_list.append(atom_type)
+    smarts_list = list(set(smarts_list))
+    extra_types = [i for i in smarts_list if i not in name_list]
+
+    for extra in extra_types:
```
ahy3nz: So in this nested-for-loop, it looks like you're looking at all the `extra_types` (superfluous atomtypes that were mentioned in smarts strings) and finding the smarts definitions that contain these extra atomtypes, then formatting the smarts definition to exclude the extra atomtype. If you found an extra atomtype in a smarts definition, then you iterate through ALL your xml-atomtype-elements to replace/modify the smarts definition.

I think one point of concern is that if you find ONE smarts definition that contains the extra type, you then modify ANY smarts definitions that contain the extra type with this new-definition, not just modifying that particular smarts definition. I think this will only work if all smarts definitions containing "extra" end up using the exact same smarts definition. 

Could you do something like:
```suggestion
    for extra in extra_types:
        for i, definition in enumerate(def_list):
            if extra in definition:
                extra_edit = '%' + extra
                extra_index = definition.find(extra_edit)
                if definition[extra_index-1] == ';':
                    new_def = definition.replace(extra_edit + ',' , '')
                else:
                    new_def = definition.replace(',' + extra_edit, '')
                atomtypes[i].set('def', new_def)
                

```
This way, you only set one smarts definition as `new_def` and not any smarts definition with `extra` inside the definition

rmatsum836:
```Also, I think test_forcefield.py has some tests for the xml_writer. I have no preference where the unit tests go, but we should try to keep the xml_writer unit tests in the same location (either all in test_forcefield or test_xml_writer)```
You're right, I didn't see these tests so I'll migrate the new ones to this file.

foyer/xml_writer.py
rmatsum836: I like this method better.

rmatsum836:
I think this is ready for review after @ahy3nz 's suggestions.  

It should be noted that this probably doesn't cover every edge case in writing out xml file snippets.  As we begin to compile a bigger collection of xml files from other groups, it would be good to further test this functionality.

ahy3nz:
Does this unit test have the case where an atomtype is listed in a smarts description, but not defined in the xml file? Or has this worked on your own systems?

rmatsum836:
This has worked on my own systems.  The system uses a force field that is still private.

mattwthompson:
@rmatsum836 could you merge master in? Travis only failing on 2.7 and we dropped it right around when you were working on this PR

rmatsum836:
Merged `upstream master` and looks like all checks are passing now.

rmatsum836:
This doesn't seem to be passing codecov.  @mwt if you're okay with this idea, I'm going to take a snippet of a IL FF xml so I can atomtype an imidazolium to try and improve the coverage.

rmatsum836:
Codecov doesn't seem to be an issue anymore.  Could we get one more reviewer on this?

mattwthompson:
Is there any way that coverage can be increased with the files already in the package, just to avoid adding files for a single test? Was it overrides that weren't being covered before?

mattwthompson:
Just for context, this is the same bug (different context but same detail) as something I found a while ago (https://github.com/ParmEd/ParmEd/issues/888) and Jason fixed (or so we thought)

mattwthompson:
So when we try to collect 1-4 scaling factors, we iterate through `struct.adjusts` and try to collect the unique factors, separated out into coulombic and LJ terms, and then return the value if there is only one unique value and error out if there is a non-1 number. That's fine for more than 1 unique value since we can't encode that much detail in our current XML format, but it doesn't work if it's zero. Here is one idea of a workaround, although the test still fails, it fails for other reasons: https://github.com/mattwthompson/foyer/commit/71df9e123d33ec36b8159c4abdae4e04340890da

ahy3nz:
> So when we try to collect 1-4 scaling factors, we iterate through `struct.adjusts` and try to collect the unique factors, separated out into coulombic and LJ terms, and then return the value if there is only one unique value and error out if there is a non-1 number. That's fine for more than 1 unique value since we can't encode that much detail in our current XML format, but it doesn't work if it's zero. Here is one idea of a workaround, although the test still fails, it fails for other reasons: [mattwthompson@71df9e1](https://github.com/mattwthompson/foyer/commit/71df9e123d33ec36b8159c4abdae4e04340890da)

Because this defaults to `1.0` for `coulomb14scale` and `lj14scale`, this is not going to be functionally the same as the original `pf6.xml` since the original had `coulomb14scale=0.5` and `lj14scale=0.5`.

(I think this is what's happening...) This is a tricky issue because this sort of thing is out of our power, as this stems from how parmed calls `struc = pmd.load_topology`. [inferring coulomb14scaling or lj14scaling](https://github.com/ParmEd/ParmEd/blob/master/parmed/openmm/topsystem.py#L413) from the `omm.System` involves looking at all the 1-4 exceptions present in the system. However..the example pf6 has no 1-4 exceptions because the bonding network doesn't stretch out to any 1-4 interactions. Without any omm 1-4 exceptions, parmed won't derive any nonbondedexceptions, coulomb14scaling, or lj14 scaling.

Maybe this means try to find a different test case that will have a 1-4 exceptions enumerated?

rmatsum836:
@ahy3nz thanks for diagnosing this.

```Is there any way that coverage can be increased with the files already in the package, just to avoid adding files for a single test? Was it overrides that weren't being covered before?```

After facing the latest setback with `pf6.xml`, I'm not sure I can test this case without adding another xml file.

I can however, use a file that is more lightweight.

rmatsum836:
I'm going to add a warning message and then should be good to go.

rmatsum836:
This should be good to go.

PR #259, opened 2019-07-23
Migrate foyer to new backend (draft 2)

ahy3nz:
Refactor atomtyping code to take backend.Topology objects, there's definitely missing functionality like atomic numbers. While a lot of it simple refactoring/variable renaming, there's some inherent missing functionality in the backend (like elements or atomic numbers) that prohibits us from utilizing all of the foyer forcefield syntax/definitions

Ignore the previous PR about backend integration.

### PR Checklist
------------


mattwthompson:
I think last time we spoke about this, we agreed it wasn't feasible to break this up into several smaller PRs? I'm fine with a big PR if it's necessary

ahy3nz:
I'm open to suggestions, but if we're trying to move from openmm and parmed, that means all the foyer operations require some refactoring to operate on a different datastructure now, which means any function that operated on `openmm.Topology` now needs to operate on `topology.Topology`, and all the syntax changes associated with that.

One option is to increase foyer's flexibility to operate on `openmm.Topology` and `topology.Topology` objects? This could definitely inflate the lines of code within foyer, but at least there's not a huge overhaul, but allowing some transition/flexibility to work with different backend datastructures

mattwthompson:
I'm actually not sure the best way to go about this. Originally I thought we'd temporarily maintain different branches for each backend but maybe having it operate on both is a better **_temporary_** solution.

PR #260, opened 2019-08-06
Change "OS" oxygen atoms in silica mol2 file to O

justinGilmer:
### PR Summary:
With the latest updates to mBuild in version 0.9.3, the atom names when
being converted to various file formats or parmed structures now undergo
a different process when their elements are being inferred.

Previously, an element name like `OS`, like in this silica.mol2 file
would evaulate to O, since `OS` is not an element.

This usually caused more issues though, because certain elements like
`SI` would be inferred as sulfur.

This now causes issues with some of our tests that rely on this silica
mol2 file, as OS is now being discovered as osmium.

The OS elements in the mol2 file have been renamed to O.

### PR Checklist
------------


mattwthompson:
If anybody were to compile a list of frustrations using biophysics software for materials simulations, the ambiguity of atom types/names/elements in mol2 files would make the cut

justinGilmer:
Thats pretty much what I did, just making sure not to override certain `TRIPOS` names to keep it as a valid mol2

ahy3nz:
> If anybody were to compile a list of frustrations using biophysics software for materials simulations, the ambiguity of atom types/names/elements in mol2 files would make the cut

"HA" and "CA" are atom names for some amino acids, then the mbuild to_parmed capitalization correction would convert to "Ha" and "Ca", then get tagged by parmed as Hahnium and Calcium, then ruin any sort of atomtyping. This probably just enforces the use of the underscore-virtual-element convention

PR #263, opened 2019-08-30
Docathon changes

mattwthompson:
### PR Summary:

Added some docstrings and other small fixes

### PR Checklist
------------

### DIFF
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -30,18 +30,19 @@ from foyer.utils.io import import_, has_mbuild
 
 
 def preprocess_forcefield_files(forcefield_files=None):
+    """Pre-process foyer Forcefield XML files"""
     if forcefield_files is None:
         return None
 
     preprocessed_files = []
 
     for xml_file in forcefield_files:
-        if not hasattr(xml_file,'read'):
+        if not hasattr(xml_file, 'read'):
             f = open(xml_file)
-            _,suffix = os.path.split(xml_file)
+            _, suffix = os.path.split(xml_file)
         else:
             f = xml_file
-            suffix=""
+            suffix = ""
 
         # read and preprocess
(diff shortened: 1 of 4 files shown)


foyer/forcefield.py
```diff
@@ -810,6 +806,7 @@ def _apply_typemap(self, topology, typemap):
             atom.id = typemap[atom.index]['atomtype']
 
     def _prepare_topology(self, topology, **kwargs):
+        """Separate positions and other topologicaly information"""
```
ahy3nz: topological*

ctk3b:
Looking like you need to adopt `black` 😁

mattwthompson:
I have been going back and forth on it for a year or two now (which inevitably means I'll go all-in sometime soon, on a whim)

ctk3b:
doit

foyer/forcefield.py
```diff
@@ -606,7 +602,7 @@ def createSystem(self, topology, nonbondedMethod=NoCutoff,
 
         data = self._SystemData
         data.atoms = list(topology.atoms())
-        for atom in data.atoms:
+        for _ in data.atoms:
```
justinGilmer: Just curious, does this work because data.atoms is a generator?

ctk3b: Are you asking about the `_` or the ability to iterate over this object?

You can iterator over it because it's an _iterable_. 

`_` is just a special python character to hold the last expression value

mattwthompson: LGTM suggested this change because the `atom` value is not used inside the loop, it could have just as well been something like `for NULL in range(len(data.atoms)):`

(although I don't understand the OpenMM API enough to follow what's happening here, I guess it's just appending num_atoms off empty lists)

PR #272, opened 2019-10-09
Mbuild imports

### LINKED ISSUE #271: Tests that require mBuild aren't marked
**Bug summary**

The following unit tests that require mBuild are not decorated with 
```python
@pytest.mark.skipif(not has_mbuild, reason="mbuild is not installed")
```

- `test_urey_bradley`
- `test_charmm_improper`
- `test_allow_empty_def`
- `test_assert_bonds`

**Code to reproduce the behavior**

Please include a code snippet that can be used to reproduce this bug.

```bash
$ pytest
```

**Software versions**

- Which version of Foyer are you using? **0.6.2**
- Which version of Python? **3.7.3**
- Which operating system? **macOS Mojave**

ahy3nz:
### PR Summary:
Fixes #271 .
Adds some pytest skips for `has_mbuild=False`

Also, before this PR, `import mbuild` as well as `mbuild.Alkanes` was located at the top of the function. This could be unsafe for testing? Although my first guess is that importing mbuild could be dangerous and circular. If we're careful in mbuild, mbuild shouldn't be importing foyer anywhere, so foyer could haphazardly import mbuild (though we should try to clean away any foyer-importing-mbuild)

### PR Checklist
------------

### DIFF
--- a/foyer/tests/test_forcefield.py
+++ b/foyer/tests/test_forcefield.py
@@ -204,6 +204,8 @@ def test_residue_map():
         assert [a0.type for a0 in b_with] == [a1.type for a1 in b_without]
         assert [a0.idx for a0 in b_with] == [a1.idx for a1 in b_without]
 
+
+@pytest.mark.skipif(not has_mbuild, reason="mbuild is not installed")
 def test_independent_residues_molecules():
     """Test to see that _check_independent_residues works for molecules."""
     from mbuild.examples import Alkane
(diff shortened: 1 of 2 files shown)


mattwthompson:
> do we want to use `mb = import_('mbuild')` instead of `import mbuild as mb`?

I _think_:
1. This is a good thing to do in principle
2. It shouldn't make a difference in these tests if the `skipif` works

so I don't have a strong opinion on it

PR #274, opened 2019-10-15
Add some style fixes suggested by LGTM

mattwthompson:
### PR Summary:
Mirrors https://github.com/mosdef-hub/mbuild/pull/616 and checks to see if LGTM behaves well here

### PR Checklist
------------

### DIFF
--- a/foyer/validator.py
+++ b/foyer/validator.py
@@ -193,5 +193,5 @@ class Validator(object):
 
 if __name__ == '__main__':
     from foyer.tests.utils import get_fn
-    v = Validator(get_fn('validate_types.xml'))
-    v = Validator(get_fn('validationerror_validate_types.xml'))
+    Validator(get_fn('validate_types.xml'))
+    Validator(get_fn('validationerror_validate_types.xml'))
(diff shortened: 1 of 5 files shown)


foyer/validator.py
```diff
@@ -193,5 +193,5 @@ def validate_overrides(self):
 
 if __name__ == '__main__':
     from foyer.tests.utils import get_fn
-    v = Validator(get_fn('validate_types.xml'))
-    v = Validator(get_fn('validationerror_validate_types.xml'))
+    Validator(get_fn('validate_types.xml'))
+    Validator(get_fn('validationerror_validate_types.xml'))
```
uppittu11: Is there a reason we made `Validator` a class? It seems like it's used just like a function with nested functions.

mattwthompson: That decision was made before my time so I can't speak to other ideas that were floated around. However, I agree it could probably be gutted in favor of its constituent functions but I'm not sure there's a downside to having it as a class.

mattwthompson:
This was mostly a test to see if LGTM behaves and so far it seems to have

PR #277, opened 2019-10-16
GAFF support in foyer

rsdefever:
### PR Summary:

Here is what I have so far for GAFF. I've included the test molecules from the MiniDrugBank under gaff_validation/minidrugbank/ante_types. These were run through antechamber to generate the GAFF atomtypes. I've included the scripts used to do that for transparency. If you run test_atomtypes.py it will create compare_antechamber-foyer.log. This shows any differences between antechamber and foyer for each molecule. Right now I have made no effort to distinguish between the conjugated pairs cc/cd, ce/cf, cg/ch, cp/cq, nc/nd, ne/nf, pc/pd, pe/pf. I have foyer label anything as the first of the pair. Details on these conjugated pairs are found on pg. 1159 of the GAFF paper (https://onlinelibrary.wiley.com/doi/epdf/10.1002/jcc.20035). Right now any molecules with Bromine or Chlorine have been moved to ante_types/removed/. These can be moved back into the main folder once we determine why parmed is reading Bromine as Boron and Chlorine as Carbon. If you examine compare_antechamber-foyer.log you will see that the atomtyping is far from perfect even when ignoring the conjugated pairs. In some cases I disagree with and/or don't understand antechambers selection. In other cases foyer is clearly wrong but I've not yet figured out a SMARTS string that fixes the problem without breaking atomtyping on another molecule. Finally, some of the SMARTS strings may be overly specific/stringent. I've been working towards making them more general and more than welcome any changes that improve the atomtyping performance.

In terms of the GAFF force field parameters, I have a script [not included] that extracts them from the gaff.dat file distributed with amber18. Right now my focus has been on developing the correct SMARTS strings. I(we) will want to confirm that all the parameters are being extracted correctly/correct unit conversions/etc prior to accepting this draft as a PR. Perhaps comparisons of some amber topologies generated by foyer and antechamber would work well for that.

Since this is all preliminary, I've created this as a draft PR. Let me know if I did anything wrong on the git side of things. 

### PR Checklist
------------


rsdefever:
Updates to GAFF implementation: Added testing on molecules used for opls validation. 100% accuracy on atom typing besides conjugated types. Updated SMARTS definitions to improve accuracy on MiniDrugBank. Added all molecules with Cl and Br to tests.

ahy3nz:
Would it be possible to move some of these files to an external repo for testing? I don't think I realized just how many test files you would be adding. It looks like you have a couple of `.py` and `.xml` files - the rest look like output, `mol2`, or `pdb`. I think the main contribution to the foyer repo would be the gaff xml, but the rest of your files might be fitting for a separate repo? That way, a user trying to clone/install foyer would only be getting the essential functionality, with the tests available elsewhere

rsdefever:
Not an issue at all; I can move it over tomorrow if that is the better approach. Based upon the opls_validation folder I was including the test cases. However, I can certainly see the argument for keeping it separate.

ahy3nz:
> Not an issue at all; I can move it over tomorrow if that is the better approach. Based upon the opls_validation folder I was including the test cases. However, I can certainly see the argument for keeping it separate.

That's a good point to bring up. I think we're aiming to move the opls_validation elsewhere, or else this repo will become over-saturated with validation files as more FFXMLs get incorporated

ahy3nz:
I would keep an eye on #282 and [this example forcefield repo](https://github.com/mosdef-hub/forcefield_perfluoroethers/pull/4) for developments. 

In theory, if someone wanted to use GAFF with foyer, they would need to install the foyer repo and install the GAFF repo. Hopefully the GAFF repo can be somewhat lightweight without having to clone all the test files? Maybe another GAFF-testing repo?

ahy3nz:
Looking at your repo via github,

* Could you add the GAFF/GAFF 2 DOIs to the atomtypes?
* Can you confirm the factors-of-2 for certain Forces? [OpenMM harmonic bonds and angles may need to be doubled](http://docs.openmm.org/latest/userguide/theory.html)
* Just for my own reference, GAFF uses lorentz-berthelot mixing rules? Your lj14 scale and coulomb14 scale look good!
* All the units/syntax look great! This is awesome thanks!

rsdefever:
Closing PR since we are moving FF XMLs into external repos and using entry points.

thangckt:
Does Foyer support GAFF now?

daico007:
Hi @thangckt, while foyer doesn't have built-in support for GAFF, the forcefield is made available through the antefoyer package (https://github.com/rsdefever/antefoyer). You can install the library per the instruction on the README or acquire the GAFF foyer XML here (https://github.com/rsdefever/antefoyer/blob/master/antefoyer/xml/gaff.xml).

thangckt:
Dear @daico007 

Thank you very much for your information. 
I wonder if `antefoyer` can assign atomic charges for a long polymer chain?
And when I use this package in a pub, then how I can cite it (I mean is there any paper of this package?)

Thank you so much

daico007:
Hi @thangckt, I don't think @rsdefever (please correct me if I am wrong) had a paper on this library yet, but I think you can still cite the github repo itself if you used it in your publication. I believe the `antefoyer` does have support for charge assignment using antechamber backend (check out `ante_charges` in https://github.com/rsdefever/antefoyer/blob/master/antefoyer/antefoyer.py). This code, though hasn't been maintained for a bit, so it may need update (and I am happy to help with what I can).

PR #278, opened 2019-10-17
Adding amide atomtype support to OPLS-AA xml file

AndrewPuente:
### PR Summary
Increased support for atomtypes related to amides in OPLS-AA. This began when I was trying to run a simulation on a box of NMP and not atomtypes could be located for one of my carbons. I added SMARTS strings, descriptions, and overrides for a total of 7 atomtypes based off of values in Jorgensen's OPLS-AA paper (10.1021/ja9621760). 

opls_237: describes the N in a primary amide such as carboxymide
opls_238: describes the N in a secondary amide such as in N-methylacetamide
opls_240: H attached to a primary amide N
opls_241: H attached to a secondary amide N
opls_242: methyl C attached to secondary amide
opls_245: describes a secondary C bonded to an amide N such as in NMP

The SMARTS strings directly reference other atomtypes when appropriate, for ex. opls_240 = [H;X1][N;%opls_237]. I ran into some issues with amide nitrogen being also identified as amine nitrogens (opls_900), so I had to add an override to opls_237 and opls_240.

For both opls_240 and opls_241 I changed their sigma values to match with the literature. They were inputted as 1.0 when Jorgensen has it as 0.0.

### PR Checklist
------------


ahy3nz:
Can you add some test molecules? 

For example [this is where we keep OPLS validation molecules](https://github.com/mosdef-hub/foyer/tree/master/foyer/opls_validation), in [the opls unit test](https://github.com/mosdef-hub/foyer/blob/master/foyer/tests/test_opls.py), each folder corresponds to a test chemical system whose atomtypes we already know. The unit test looks through these folders, gets the atomic structure/bonding/elements and attempts to identify atom types for these compounds, and then we compare to the "answer key" of correct atom types within the `top` file.

So could you build some small molecules whose atomtypes you *already* know, and see if foyer's atomtyping will generate the correct atomtypes?

mattwthompson:
@AndrewPuente anything we can do to help move this along?

You may also want to merge upstream changes into your feature branch:

```
git fetch upstream
git checkout master
git merge upstream/master
git checkout amide_atomtypes_oplsaa
git merge master
```

foyer/tests/implemented_opls_tests.txt
```diff
@@ -35,22 +36,29 @@ diethylene-glycol
...
 isobutane
+methane
 methanol
 n-butylamine
+N-methylacetamide
+neopentane
 nitrobenzene
 nitroethane
 nitromethane
+nmp
```
mattwthompson: Please use the full name in case something else may be interpreted as `nmp` in the future

AndrewPuente:
Thanks Alex! I went ahead and alphabetized all the most recent additions to the implemented_opls_tests.txt file.

I also suspect that opls_244 refers to the carbon in RCH2 that is attached to an secondary amide nitrogen. However, this atomtype and the corresponding hydrogens are not mentioned in the original OPLS paper, so I left those out of this PR.

foyer/tests/implemented_opls_tests.txt
AndrewPuente: On it! I'm used to it as NMP but better safe than sorry. I'm changing the appropriate files to N-methylpyrrolidone.

I didn't have a reason to change it from GRO/TOP to MOL2, per say. I just made those mol2 files from scratch and double-checked the atomtypes. I also wasn't aware of the issue with ParmEd conversions but I did not convert those GRO/TOP  files directly to MOL2. Do you think it would be better to leave them in the original GRO/TOP?

Issue #279, opened 2019-10-17
NetworkX latest update causes issues


justinGilmer:
**Bug summary**
Was reviewing [this PR in mBuild](https://github.com/mosdef-hub/mbuild/pull/618) and noticed a lot of failures in the tests.

Turns out it was due to Foyer and the [SMARTS Graph](https://travis-ci.org/mosdef-hub/mbuild/jobs/599163797).

[NetworkX recently updated to 2.4 and  the way we access nodes have been removed: `G.node –> use G.nodes`](https://networkx.github.io/documentation/stable/release/release_2.4.html)

**Code to reproduce the behavior**

Updated to networkX 2.4 and run foyer tests that use SMARTS graphs.

It appears that we just have to change from `G.node` to `G.nodes`.

uppittu11:
Good catch! I was wondering why CI was failing

mikemhenry:
Lets keep this issue open even post PR until we get the networkx pin sorted out.

jennyfothergill:
has this pin been sorted out?

justinGilmer:
It has!

PR #280, opened 2019-10-17
Update atomtyper.py to use networkX 2.X methods

justinGilmer:
### PR Summary:
Upgrading to networkX >= 2.4 removes the nx.Graph.node method, breaking
atomtyper.py.

NetworkX recently updated to 2.4 and the way we access nodes have been
removed: G.node –> use G.nodes. This was only present in `atomtyper.py`,
the rest of our node calls were using the new syntax.

Updating to handle the new syntax restores expected behavior, and all
unit tests pass.

### PR Checklist
------------

Addresses issue #279

### DIFF
--- a/foyer/atomtyper.py
+++ b/foyer/atomtyper.py
@@ -26,7 +26,7 @@ def find_atomtypes(topology, forcefield, max_iter=10):
     subrules = dict()
     system_elements = {a.element.symbol for a in topology.atoms()}
     for key,val in rules.items():
-        atom = val.node[0]['atom']
+        atom = val.nodes[0]['atom']
         if len(list(atom.find_data('atom_symbol'))) == 1 and \
                     not list(atom.find_data('not_expression')):
             try:


mikemhenry:
Since we didn't pin the last version, we should probably consider making this a hotfix and re-doing the release since anyone downloading the current version off of conda will run into problems... I think...

justinGilmer:
I agree, any updates and new installs will not work with our current releases

justinGilmer:
I am not sure if that is necessary. We already pin to >= 2.0, which had the `graph.nodes` method. The issue was we were using a deprecated method from 1.X. What benefit would pinning the version bring?

mikemhenry:
It would mean that if we had pined to to <2.4, things wouldn't have broken. We should be pining most of our stuff like that IMHO since we don't want a change upstream breaking a release. When we are ready to do a release we could relax the pinning, see what versions are working, then re-pin on those to keep things updated.

justinGilmer:
Ok, I think I might have miss-represented what happened here. `nx.graph.nodes` has been available since version 2.0, and works for all versions of NX >=2.0. `nx.node` was deprecated since 2.0, we just never made the change in syntax. So this update should be backwards compatible with all versions of 2.X. 

The big breaking changes were fixed when pinning to >=2.0, we just ignored the deprecation warning it provided. maybe pinning NX >=2.0 <=3.0 might be more correct?

mikemhenry:
That sounds reasonable, it was our fault for not updating it before it was broken... We should consider running CI with fail on warn or something to force us to fix? Either way I'm happy with pinning it the way you described.

mattwthompson:
I'm of two minds on pinning; in general I think over-pinning is something we want to avoid, but I don't see how we can insulate ourselves against breaking upstream changes unless we catch issues like this and push out hotfix releases on a short lead time (like a day). In this case, since this new syntax works on both new and older versions, I think this is a good change. But we should have a larger discussion about pinning and how to manage our deps.

mikemhenry:
I don't mind being overruled in this situation since its 2v1 right now, but it looks like 2.5 isn't even out yet, so pinning less than version 3 should be fine.

mattwthompson:
You did bring up a good point when you suggested when push out a hotfix for this change. This seems necessarily and fairly urgent, no?

mikemhenry:
Yes, anyone who installs foyer will run into an issue. Once we get this merged people who update from source will be okay, but until we re-release, it will be broken for people. All previous versions that didn't have the pin to the right networkx version will stay broken unless a user downgrades to the right version of networkx.

uppittu11:
I couldn't find any `networkx` deprecation warnings in the CI build logs for `mbuild` or `foyer`. I'm also not getting any warnings if I use `Graph.node` using `networkx=2.3` locally. I'm not sure how we could have caught this issue before now.

mikemhenry:
`python -c "import warnings; print(warnings.filters)"` will show current filters:
```
[('ignore', None, <class 'DeprecationWarning'>, None, 0), 
('ignore', None, <class 'PendingDeprecationWarning'>, None, 0), 
('ignore', None, <class 'ImportWarning'>, None, 0), 
('ignore', None, <class 'BytesWarning'>, None, 0), 
('ignore', None, <class 'ResourceWarning'>, None, 0)]
```
So unless you use something like `python -Wall` you won't see them

mattwthompson:
Not sure if this is simple but it would be nice to have some analytics from AZP about warnings, if that can be parsed through pytest to the web interface. Maybe there's a plugin for that?

Issue #187, opened 2018-11-06
Maintaining different versions of force fields


mattwthompson:
Not sure we've talked about this much, but I think it's worth recording somewhere.

Should version control be handled within `foyer`, in any capacity, or should it be done purely externally, i.e. with git?

Hypothetically (read: something I need done for my own research) I have a force field with recently-updated parameters, but it is useful to have an old version for comparisons. It happens to be buried in the git history, but it's not so easy to grab an old version of a force field just using git. It's not like we have `myff = Forcefield('myff.xml', commit='blah123')`. Grabbing an older version of the file and using just using `myff_old.xml` would work for my purposes, but it doesn't seem to be preferred.

ctk3b:
> Should version control be handled within foyer?

What would you envision `foyer` providing here? At some level you need the copies of the files lying around somewhere.

Off the top of my head you could:
1) have file names that contain the version
2) use git release tags to identify "released" versions

I would probably look to how OpenMM packages their set of supported forcefields and how the OpenFF effort has released versions of their frosst FF

ahy3nz:
If you did something like `myff = Forcefield('myff.xml', commit='blah123')`, doesn't that assume a particular git repo for that force field? What if you tried pulling the xml from the URL of a particular git commit, loading the xml, and then deleting it? That way there's room for the user to specify which repo and which commit to pull the file
```
import requests
url = 'https://raw.githubusercontent.com/choderalab/openmm-forcefields/f94144ea1b9604c06c9831c96311c94422b3179a/charmm/ffxml/charmm36.xml'                                    r = requests.get(url, allow_redirects=True)
open('tempff.xml', 'wb').write(r.content)
```

mattwthompson:
After speaking with @summeraz and @justinGilmer what I have in mind is a way for a user to access old versions of the force field that relies on some sort of metadata encoded in the xml files, maybe like a more useful approach than #153. This definitely could all be handled with git (e.g. maintaining a branch that includes old versions of a force field, or simply using releases), but this would not be accessible to a user who only installs the package from conda and not a development version. It doesn't look like openmm has a default way of handling this, instead just carrying around [different files ](https://github.com/pandegroup/openmm/tree/master/wrappers/python/simtk/openmm/app/data)for older versions. - it seems like their large force fields are already tied together into [single major](
[...]

PR #282, opened 2019-10-22
Establish entry point for external force fields

mattwthompson:
### PR Summary:

This is my first stab at setting up foyer to interface with external force fields via `entry_points`. This works analogously to [recipes in mbuild](https://github.com/mosdef-hub/mbuild/blob/63a9bbae6c5f2c2bbc55d8f1738b30dfb896aeed/mbuild/recipes/__init__.py) and similarly to [a force field](https://github.com/openforcefield/smirnoff99Frosst) distributed by the openforcefield folks. The important point to repeat is that, if it is package is set up the right way, loading in an external force field to foyer does not require further modification to the foyer source code. It follows that external users can build a conda package, upload it to the cloud, and then it's there for anybody who wants to use the force field (again, without adding anything here).

This would look something like this to a user (using https://github.com/mattwthompson/OPLSAA-foyer):

```
$ conda install -c some-channel oplsaafoyer
```
```python3
>>> import foyer
>>> my_ff = foyer.external_forcefields.get_ff_path()[0] + '/oplsaa.xml'
>>> OPLSAA = foyer.Forcefield(my_ff)
/Users/mwt/software/foyer/foyer/validator.py:132: ValidationWarning: You have empty smart definition(s)
  warn("You have empty smart definition(s)", ValidationWarning)
<foyer.forcefield.Forcefield object at 0x10894d0b8>
# Can proceed from here identically to as if we loaded in the XML from inside foyer ...
```

There has been a lot of discussion about this sort of feature while writing the grant but I believe the only related discussion here has been #187 

What do we think @mosdef-hub/mosdef-contributors?

### PR Checklist
------------


uppittu11:
This is pretty awesome! Just tested out and loading the forcefield and it works. 

A couple of questions before I approve this: 
1. So if you try to install multiple external forcefields, would `foyer.external_forcefields.__init__` and `get_ff_path()` be clobbered? Can you only have one external forcefield installed at a time?
2. Your idea in #187 discussed forcefield versions. Is there a way to access the `version` listed in `setup.py`? Should we add a `version` or a `name` attribute to foyer forcefields? You could probably do this by actually loading the forcefield as a subclass of `Forcefield` in `oplsfoyer.py` (kind of like what we do for mbuild recipes), and then add a `version` attribute to that subclass.

mattwthompson:
1. Good point. I'm not sure how to implement this but the ideal behavior would be for them to each have similar-looking `get_something()` functions but append paths instead of clobbering them.

2. I was thinking more along the lines of using conda to manage [versions](https://github.com/openforcefield/smirnoff99Frosst/releases) of a force fields - maybe we could also tag the version and/or name but use it only as metadata?

mattwthompson:
I consider this ready for review; not much needs to be done on the foyer side of thing. There are some decisions to be made on how an external force field "package" should interact with it, see a few examples in https://github.com/mosdef-hub/forcefield_perfluoroethers/pull/4

ahy3nz:
Just summarizing some discussion with others, maybe some room to re-organize the entry points:
As it stands now, external force field are accessed via `foyer.external_forcefields.PFE` will return an instance of `foyer.Forcefield`. 

Maybe the entry point should be `foyer.forcefields.PFE`, `foyer.forcefields.OPLSAA`, etc. that will return instances of `foyer.Forcefield`, or functions to return the path-to-XML `foyer.forcefields.get_perfluoroether_path()`. And if we still include any forcefields/FFXMLs in foyer, store those forcefields in `foyer.lib.forcefields`, but still accessible via entrypoints in `foyer.forcefields.X`

The next discussion point, if I recall, was versioning/storing the XMLs. OFF stores all XML versions in the HEAD/latest commit, but tagging commits and Zenodo DOIs as they go. Regardless, you could be on the latest OFF commit, but still access an old XML without checking out a new tag. So you wouldn't have to check out a particular commit/tag/release to access an old XML. The advantage is that you don't have to worry about checking out a commit - all the files are at your disposal. The disadvantage is precisely that, with all the files at your disposal, how does `foyer.forcefields.PFE` know which `PFE-ver.x.y.z.xml` to parse?

mattwthompson:
After the last changes (and my stealthy rebase/amend/force-push maneuver), this is how it looks to the user:

```python3
>>> import foyer
/Users/mwt/software/foyer/foyer/validator.py:132: ValidationWarning: You have empty smart definition(s)
  warn("You have empty smart definition(s)", ValidationWarning)
/Users/mwt/software/foyer/foyer/forcefield.py:395: UserWarning: Non-atomistic element type detected. Creating custom element for _CH4
  'Creating custom element for {}'.format(element))
/Users/mwt/software/foyer/foyer/forcefield.py:395: UserWarning: Non-atomistic element type detected. Creating custom element for _CH3
  'Creating custom element for {}'.format(element))
/Users/mwt/software/foyer/foyer/forcefield.py:395: UserWarning: Non-atomistic element type detected. Creating custom element for _CH2
  'Creating custom element for {}'.format(element))
/Users/mwt/software/foyer/foyer/forcefield.py:395: UserWarning: Non-atomistic element type detected. Creating custom element for _HC
  'Creating custom element for {}'.format(element))
>>> foyer.forcefields.OPLSAA
<foyer.forcefield.Forcefield object at 0x11b6aff98>
>>> foyer.forcefields.TRAPPE_UA
<foyer.forcefield.Forcefield object at 0x11b859c18>
>>> foyer.forcefields.PFE
<foyer.forcefield.Forcefield object at 0x11ba466a0>
```

Note that the last one is an "external" plugin developed in another repo. There are some other functions that return paths to globs or XMLs if a user really wants them, but these capitalized objects are the force fields themselves, loaded into memory. Since `foyer/forcefields/` is now a module I moved the XMLs into a new folder so as to separate out the Python files from the data.

ahy3nz:
Random thought, I wonder if it's dangerous to have `foyer.Forcefields.PFE` point to a specific *instance* as opposed to the module/class (that hasn't been instantiated). If I chose to modify some random attribute of `foyer.Forcefields.PFE.some_random_attribute`, that is going to affect the state of any subsequent calls to `foyer.Forcefields.PFE` since all these calls are referring to the same object. 

I guess the response to such an involved-use of the API should mean the user should use some other API functionality like `foyer.Forcefield(foyer.Forcefield.get_perfluoroether_path())`, where we are no longer dealing with the same forcefield instance/object, but rather creating a new one every time.

mattwthompson:
That's an interesting point and something we should consider when we implement versioning and if the forcefield class evolves in the future. However, right now we don't really have any user-facing attributes, I don't think?

Merging for now so we can get external packages rolling. Hopefully the location of the entry point doesn't change and we only add new features to it.

PR #283, opened 2019-10-24
Mock requests in tests

### LINKED ISSUE #211: Mark some tests as flaky

mattwthompson:
### PR Summary:
Fixes #211, unlike my crash of an attempt in #212

CI might take some negotiation
 
### PR Checklist
------------

### DIFF
--- a/foyer/tests/test_forcefield.py
+++ b/foyer/tests/test_forcefield.py
@@ -12,12 +12,26 @@ from foyer import Forcefield
 from foyer.forcefield import generate_topology
 from foyer.forcefield import _check_independent_residues
 from foyer.exceptions import FoyerError, ValidationWarning
-from foyer.tests.utils import get_fn
+from foyer.tests.utils import get_fn, register_mock_request
 from foyer.utils.io import has_mbuild
 
 
 FF_DIR = resource_filename('foyer', 'forcefields')
 FORCEFIELDS = glob.glob(os.path.join(FF_DIR, '*.xml'))
+
+RESPONSE_BIB_ETHANE_SINGLE = """@article{Jorgensen_1996,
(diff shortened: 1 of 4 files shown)


foyer/tests/test_forcefield.py
```diff
@@ -105,14 +149,6 @@ def test_write_refs_multiple():
                                              n=0))
     assert not diff
 
-@pytest.mark.skipif(not has_mbuild, reason="mbuild is not installed")
-def test_write_bad_ref():
```
mattwthompson: Any reason to have dropped this test?

umesh-timalsina: We are already faking the original response, so, it made sense since a point of failure would be the `request`. We could use same tricks to generate user warnings. I have no opinion one way or another.

mattwthompson: I misunderstood this test at first, I think it's fine to remove it.

umesh-timalsina:
@uppittu11 We can do a better job of covering the lines. 3cf32df adds a 404 (Not Found) mock to the `test_write_bad_ref` which was removed. Should do the trick for `codecov`.

mattwthompson:
Thanks @umesh-timalsina!

PR #288, opened 2019-10-30
Switch entrypoints to functions that load classes

mattwthompson:
### PR Summary:
Small change post-#282 per suggestions of @uppittu11 and @ahy3nz https://github.com/mosdef-hub/foyer/pull/282#issuecomment-547144491 https://github.com/mosdef-hub/forcefield_perfluoroethers/pull/4/files#r339807274
Instead of the entrypoint being an instance of the force field, it's a function that grabs it. This cleans up the import such that the user more discretely instantiates the object. It also prevents the loading from happening upon `import foyer` and should save some time on imports. A quick local check saved ~0.3 seconds, no promises this is accurate.

### PR Checklist
------------

### DIFF
--- a/foyer/tests/test_plugin.py
+++ b/foyer/tests/test_plugin.py
@@ -6,9 +6,9 @@ def test_basic_import():
     assert 'forcefields' in dir(foyer)
 
 
-@pytest.mark.parametrize('ff_name', ['OPLSAA', 'TRAPPE_UA'])
-def test_forcefields_exist(ff_name):
-    ff_name in dir(foyer.forcefields)
+@pytest.mark.parametrize('ff_loader', ['load_OPLSAA', 'load_TRAPPE_UA'])
+def test_forcefields_exist(ff_loader):
+    assert ff_loader in dir(foyer.forcefields)
 
 
 def test_load_forcefield():
--- a/foyer/tests/test_plugin.py
+++ b/foyer/tests/test_plugin.py
@@ -6,9 +6,11 @@ def test_basic_import():
     assert 'forcefields' in dir(foyer)
 
 
-@pytest.mark.parametrize('ff_loader', ['load_OPLSAA', 'load_TRAPPE_UA'])
-def test_forcefields_exist(ff_loader):
-    assert ff_loader in dir(foyer.forcefields)
+def test_loading_forcefields():
+    #funcs = [func for func in dir(foyer.forcefields) if 'load' in func and '__' not in func]
+    for func in dir(foyer.forcefields):
+        if 'load_' in func and '__' not in func:
+            eval('foyer.forcefields.' + func)()
 
 
 def test_load_forcefield():
--- a/foyer/forcefields/forcefields.py
+++ b/foyer/forcefields/forcefields.py
@@ -27,5 +27,5 @@ def get_forcefield(name=None):
                 ' in path {}'.format(name, get_ff_path()))
     return Forcefield(ff_path)
 
-OPLSAA = get_forcefield(name='oplsaa')
-TRAPPE_UA = get_forcefield(name='trappe-ua')
+load_OPLSAA = get_forcefield(name='oplsaa')
+load_TRAPPE_UA = get_forcefield(name='trappe-ua')
--- a/foyer/forcefields/forcefields.py
+++ b/foyer/forcefields/forcefields.py
@@ -27,5 +27,14 @@ def get_forcefield(name=None):
                 ' in path {}'.format(name, get_ff_path()))
     return Forcefield(ff_path)
 
-load_OPLSAA = get_forcefield(name='oplsaa')
-load_TRAPPE_UA = get_forcefield(name='trappe-ua')
+
+def load_OPLSAA():
+    return get_forcefield(name='oplsaa')
+
+
+def load_TRAPPE_UA():
+    return get_forcefield(name='trappe-ua')
+
+
+load_OPLSAA = load_OPLSAA
+load_TRAPPE_UA = load_TRAPPE_UA
--- a/foyer/forcefields/forcefields.py
+++ b/foyer/forcefields/forcefields.py
@@ -34,7 +34,3 @@ def load_OPLSAA():
 
 def load_TRAPPE_UA():
     return get_forcefield(name='trappe-ua')
-
-
-load_OPLSAA = load_OPLSAA
-load_TRAPPE_UA = load_TRAPPE_UA
--- a/foyer/tests/test_validator.py
+++ b/foyer/tests/test_validator.py
@@ -17,7 +17,7 @@ ERRORS = {'validationerror': (ValidationError, MultipleValidationError),
 }
 
 FF_DIR = resource_filename('foyer', 'forcefields')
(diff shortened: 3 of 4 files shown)


uppittu11:
Looks like the last test is being skipped now (it wasn't before). Should we drop it or fix it?

```bash
foyer/tests/test_validator.py::test_forcefields[ff_file0] SKIPPED
```

https://travis-ci.org/mosdef-hub/foyer/jobs/605218247?utm_medium=notification&utm_source=github_status

foyer/tests/test_plugin.py
```diff
@@ -6,9 +6,9 @@ def test_basic_import():
     assert 'forcefields' in dir(foyer)
 
 
-@pytest.mark.parametrize('ff_name', ['OPLSAA', 'TRAPPE_UA'])
-def test_forcefields_exist(ff_name):
-    ff_name in dir(foyer.forcefields)
+@pytest.mark.parametrize('ff_loader', ['load_OPLSAA', 'load_TRAPPE_UA'])
+def test_forcefields_exist(ff_loader):
+    assert ff_loader in dir(foyer.forcefields)
```
uppittu11: What's the purpose of this test? I thought we are trying to move opls and trappe to their own repos, like PFE. 

If we could loop through *all* of the installed plugins somehow, this test would be more all-inclusive and make more sense to have. Try something like the snippet below. Or if there is a cleaner/better way of getting these functions. 

```python3
@pytest.mark.parameterize('ff_loader', 
        [func for func in dir(foyer.forcefields) 
         if 'load' in func and '__' not in func])
```

mattwthompson: > What's the purpose of this test?

It's just to make sure those loader functions are there, which ...

> I thought we are trying to move opls and trappe to their own repos, like PFE.

assumes, for now that we are shipping these force fields with the main repo.

I like your idea for iterating through the loaders that are in the entry point group gets at what I was going for but isn't this self-assuring? Since it only iterates through the loaders it finds, is there even a way for it to fail? I guess it could fail if the loader is bad, but the point was to check to see the loaders that we hope are there indeed exist.

mattwthompson:
>  We should remove this test because there are no forcefield xmls in foyer/forcefields/.

I put them in `foyer/forcefields/xml/` but didn't update the glob in that path

uppittu11:
Maybe we could have a function that detects which plugins are installed when you import foyer (it could save/print the list of plugins too). And this can be used to iterate through all the plugins for this test. We can stick this in another PR if that’s appropriate.

On Oct 30, 2019, at 7:20 PM, Matt Thompson <notifications@github.com> wrote:

﻿

@mattwthompson commented on this pull request.

________________________________

In foyer/tests/test_plugin.py<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmosdef-hub%2Ffoyer%2Fpull%2F288%23discussion_r340915935&data=02%7C01%7Cp.shama%40vanderbilt.edu%7C94ec9de2f6004448d14108d75d981e67%7Cba5a7f39e3be4ab3b45067fa80faecad%7C0%7C0%7C637080780257946600&sdata=nl%2Fm2ArV9W%2FIOC7wpOHS40UGG3qqr0BSAXYmfnvOOWM%3D&reserved=0>:

> @@ -6,9 +6,9 @@ def test_basic_import():
     assert 'forcefields' in dir(foyer)


-@pytest.mark.parametrize('ff_name', ['OPLSAA', 'TRAPPE_UA'])
-def test_forcefields_exist(ff_name):
-    ff_name in dir(foyer.forcefields)
+@pytest.mark.parametrize('ff_loader', ['load_OPLSAA', 'load_TRAPPE_UA'])
+def test_forcefields_exist(ff_loader):
+    assert ff_loader in dir(foyer.forcefields)


What's the purpose of this test?

It's just to make sure those loader functions are there, which ...

I thought we are trying to move opls and trappe to their own repos, like PFE.

assumes, for now that we are shipping these force fields with the main repo.

I like your idea for iterating through the loaders that are in the entry point group gets at what I was going for but isn't this self-assuring? Since it only iterates through the loaders it finds, is there even a way for it to fail? I guess it could fail if the loader is bad, but the point was to check to see the loaders that we hope are there indeed exist.

mattwthompson:
```python3
>>> funcs = [func for func in dir(foyer.forcefields) if 'load' in func and '__' not in func]
>>> funcs
['load_OPLSAA', 'load_TRAPPE_UA']
>>> [eval('foyer.forcefields.' + func)() for func in funcs]
[<foyer.forcefield.Forcefield object at 0x11ab12470>, <foyer.forcefield.Forcefield object at 0x11a89aa58>]
```

I think this accomplishes what you suggested. I also noticed that even though we're checking these functions exist, we're not checking to see that they work. Above should fix that, I think.

https://codecov.io/gh/mosdef-hub/foyer/src/update-entrypoints/foyer/forcefields/forcefields.py

foyer/tests/test_plugin.py
uppittu11: The changes you made now look fine to me, but looking to the future we could probably change a few things: 

At this point the entry point is working on foyer's end. So I think foyer's responsibility should only be to detect whatever plugins in can find in `foyer.forcefields`. The burden is on the maintainers of the specific forcefield plugin repo to make sure their forcefield is detected by the foyer (i.e. they put the correct entry point location or whatever) and that should be checked with their own tests in their own CI. 

So, in my mind, the tests in foyer should only test the forcefields it detects, it shouldn't test if a forcefield is detected correctly.

That being said we could develop a [more sophisticated plugin detector](https://github.com/mosdef-hub/foyer/pull/288#issuecomment-548465083).

mattwthompson: > The burden is on the maintainers of the specific forcefield plugin repo to make sure their forcefield is detected by the foyer 

That's a good point, I agree

foyer/tests/test_plugin.py
```diff
@@ -6,9 +6,11 @@ def test_basic_import():
...
 
 
-@pytest.mark.parametrize('ff_name', ['OPLSAA', 'TRAPPE_UA'])
-def test_forcefields_exist(ff_name):
-    ff_name in dir(foyer.forcefields)
+def test_loading_forcefields():
+    #funcs = [func for func in dir(foyer.forcefields) if 'load' in func and '__' not in func]
+    for func in dir(foyer.forcefields):
+        if 'load_' in func and '__' not in func:
+            eval('foyer.forcefields.' + func)()
```
justinGilmer: Im ok with this, but just want to note that eventually this might be a possible issue. However, I cannot think of a better way to handle this.

mattwthompson: I agree, `eval` is dangerous - could you raise an issue so we don't forget?

ahy3nz:
@rsdefever try to update gaff with these functions

uppittu11:
> @rsdefever try to update gaff with these functions

Check out this PR too for inspiration. https://github.com/mosdef-hub/forcefield_perfluoroethers/pull/7

rsdefever:
@ahy3nz @uppittu11 take a look at my GAFF-foyer repo. I think it is now working with the latest entry points in foyer.

Issue #290, opened 2019-10-31
Plugin Detector


uppittu11:
**Describe the behavior you would like added to Foyer**
A plugin detector within foyer that will detect all of the `foyer` plugins installed in the environment. We would also want convenient way to access all of the information about a plugin. See #288.

**Additional context**

> In my view, ideally we would have a global `dict` of the installed plugins, so then we can call a variable (`foyer.plugins` or something like that) and it outputs the `dict` of the installed plugins and all of their info (like version, pointer to loader function, and location of xml, etc). 
> 
> ```python3
> >>> import foyer
> >>> foyer.plugins
> {
> "oplsaa" : {"version" : "0.0.3", "loader_function" : foyer.forcefields.load_OPLSAA, "path" : "/path/to/oplsaa.xml"}, 
> "trappe-ua" : {"version" : "0.0.5", "loader_function" : foyer.forcefields.load_TRAPPE_UA, "path" : "/path/to/trappe_ua.xml"}, 
> "perfluoroethers" : {"version" : "loader_function" : "0.0.1", foyer.forcefields.load_PFE, "path" : "/path/to/perfluoroethers.xml"}
> }
> ```
> 
> This way we can easily check what forcefields are installed and these would be the only forcefields tested.
> 
> _Originally posted by @uppittu11 in https://github.com/mosdef-hub/foyer/pull/288#issuecomment-548465083_
>

PR #292, opened 2019-11-01
Add initial attempt at a plugin detector.

uppittu11:
Here is an attempt at a writing a plugin detector (#290). See `foyer.plugins:collect_plugins()`

Functions named `foyer.forcefields.load_{plugin_name}` are detected as plugins, but there is also a `plugin_names` argument to supply a list of plugin function names that don't fit that mold. The `version`, `xml_file`, and `load_function are collected into a `dict` and returned. 

I still don't know how to get the `version`. Maybe we could add it as an attribute of the plugin forcefield object when we load it up:

```python3
def load_OPLSAA():
    ff = get_forcefield(name='oplsaa')
    ff.version = 0.0.1
    return ff
```

### DIFF
--- a/foyer/plugins.py
+++ b/foyer/plugins.py
@@ -1,0 +1,39 @@
+import foyer
+import glob
+
+def collect_plugins(plugin_names=None):
+    """
+    Detects which plugins are installed
+
+    Arguments
+    ---------
+    plugin_names: list, default=None
+        A list of non-standard plugin names (strings) to additionally
+        check for.
+    """
+
+    plugin_funcs = [func for func in dir(foyer.forcefields)
+                    if "load" in func and "__" not in func]
+
+    if plugin_names:
+        plugin_funcs += plugin_names
+
+    plugins = dict()
+    for plugin_func in plugin_funcs:
+        plugin_loader_func = eval("foyer.forcefields.{}".format(plugin_func))
+        # Assumes all plugins are named "load_{plugin_name}"
+        plugin_name = "_".join(plugin_func.split("_")[1:])
+        # TODO: plugin_version = get_version_info
+        plugin_dir = eval("foyer.forcefields.{}.__globals__['__file__']".format(plugin_func))
+        # TODO: plugin_xml_path = get_xml_path
+        # This assumes that plugin directory tree is consistent.
+        # Does not consider versioned FFs.
+        plugin_xml_path = glob.glob("{}/xml/*xml".format(plugin_dir))
+
+        plugin = {plugin_name : {"version"       : None,
+                                 "load_function" : plugin_loader_func,
+                                 "xml_path"      : plugin_xml_path
+                                 }}
+        plugins.update(plugin)
+
+    return plugins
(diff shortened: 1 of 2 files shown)


foyer/plugins.py
```diff
@@ -0,0 +1,39 @@
...
+
+    if plugin_names:
+        plugin_funcs += plugin_names
+
+    plugins = dict()
+    for plugin_func in plugin_funcs:
+        plugin_loader_func = eval("foyer.forcefields.{}".format(plugin_func))
+        # Assumes all plugins are named "load_{plugin_name}"
+        plugin_name = "_".join(plugin_func.split("_")[1:])
+        # TODO: plugin_version = get_version_info
```
mattwthompson: Any initial thoughts on how we should store this information? I see we want it to be accessible from the loader but I'm not sure how to go about doing that - maybe these functions should be expanded into (lightweight) classes?

foyer/plugins.py
```diff
@@ -0,0 +1,39 @@
...
+        plugin_dir = eval("foyer.forcefields.{}.__globals__['__file__']".format(plugin_func))
+        # TODO: plugin_xml_path = get_xml_path
+        # This assumes that plugin directory tree is consistent.
+        # Does not consider versioned FFs.
+        plugin_xml_path = glob.glob("{}/xml/*xml".format(plugin_dir))
```
mattwthompson: Is there a way to semi-intelligently glob recursively, i.e. to get both `'{}/xml/*xml'` and `'{}/*xml'`? We probably don't need arbitrary recursion, maybe handling 2-3 cases is sufficient.

foyer/plugins.py
uppittu11: Check out my [plugins branch for the PFE repo](https://github.com/uppittu11/forcefield_perfluoroethers/tree/plugins). To demonstrate how FF versioning would work, I added some dummy forcefield versions and file organization for adding forcefields versions. Version and xml_path are returned when the forcefield is loaded, and can be easily accessed by the plugin detector in foyer:

```python3
>>> import foyer
>>> pfe = foyer.forcefields.load_PFE()
>>> pfe.version
'1.0.0'
>>> pfe.xml_path
['/Users/parashara/Documents/devel/git/forcefield_perfluoroethers/perfluoroethers/xml/current/PFE.xml']
>>> pfe = foyer.forcefields.load_PFE("0.0.2")
>>> pfe.version
'0.0.2'
>>> pfe.xml_path
['/Users/parashara/Documents/devel/git/forcefield_perfluoroethers/perfluoroethers/xml/versions/0.0.2/PFE.xml']
```

foyer/plugins.py
rmatsum836: For me this is what I get:

```
(Pdb) plugin_dir
'/Users/raymatsumoto/science/forcefields/forcefield_perfluoroethers/perfluoroethers/perfluoroethers.py'
(Pdb) 
```
So I'm unable to glob this path.

uppittu11:
Ideally all of this information (like the xml file location and version) will be stored in the forcefield class, so we don't have to do any globbing. I think this is something @mattwthompson is working on

PR #293, opened 2019-11-01
`_check_angles` bug fix

daico007:
### PR Summary:
Reorder the codes for `_check_angles`. Right now, even if `verbose=True`, the corresponding block of code will not be excuted because of the `_error_or_warn` line right before it. 
### PR Checklist
------------

### DIFF
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -263,12 +263,6 @@ def _check_bonds(data, structure, assert_bond_params):
 
 def _check_angles(data, structure, verbose, assert_angle_params):
     """Check if all angles were found and parametrized."""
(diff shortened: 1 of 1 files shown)


daico007:
The code will continue if only the warning is raised. 
Just look back at the code. So before, it seems like the `if verbose:` code is only activated if we pass in `(assert_angles_params=False, verbose=True)`. So if we only pass in `verbose=True` without the `assert=False`, the code will error out without giving the missing parameters.
So I guess the question becomes whether we want to keep the old behavior (need to pass in both args) or just pass in `verbose=True`, the code will print out the missing params and then error out.

PR #296, opened 2019-11-10
Remove OpenMM from atom-typing step

mattwthompson:
### PR Summary:

Work in progress idea built off of #294, but takes it a step farther and modifies `run_atomtyping` to operate on ParmEd structures, not OpenMM. This is a step in the direction of making things more modular. In principle, atom-typing does not require any OpenMM stuff; just enough information to construct a `SMARTSGraph`. The changes here (provided I can get them to work) will still use OpenMM for the parametrization step.

Right now, this works for the simplest cases but doesn't yet work for non-element types and residue maps.

### PR Checklist
------------


ahy3nz:
Addressing custom element functionality would address a plurality of the test fails, I think

rsdefever:
@mattwthompson I like these changes. Let me know if I can help finish them off.

foyer/forcefield.py
```diff
@@ -125,6 +125,7 @@ def _topology_from_parmed(structure, non_element_types):
                 element = elem.Element.getBySymbol(pmd_atom.name)
 
         omm_atom = topology.addAtom(name, element, residues[pmd_atom.residue])
+        omm_atom.id = pmd_atom.id
```
rmatsum836: Per our conversation earlier, I think `omm_atom.id` refers to the atom type.  a ParmEd `Atom` doesn't have an `id` attribute and we should be calling `type` instead.

foyer/smarts_graph.py
```diff
@@ -215,7 +216,7 @@ def __init__(self, G1, G2, node_match, element, typemap):
         if element not in [None, '*']:
             self.valid_nodes = [n for n, atom in 
                                 nx.get_node_attributes(G1, name='atom').items()
-                                if atom.element.symbol == element]
+                                if atom.element == md.element.get_by_symbol(element).atomic_number]
```
justinGilmer: This might already be taken care of since it is a parmed structure, but this could run into the same issues as before with improperly capitalized elements. 
```suggestion
                                if atom.element == md.element.get_by_symbol(element.capitalize()).atomic_number]
```

mattwthompson:
Fasten your seatbelts, this turned out to be much hairier than I originally hoped ...

So the main driver in the `Forcefield` class is the `apply` method. It does two things: find atom types and then, based on those atom types, parametrize the system. As defined here, finding the atom types is simply solving the atom-typing problem and tagging atom types, i.e. the only deliverable is a string representing the name of the atom type. The second step involves actually storing the force field information in some object (LJ parameters, charges, picking the right bonded potentials, and also encoding the bonded parameters, etc.). After #294, these steps are handled by two discrete methods, `run_atomtyping` and `parametrize_system`. The latter is mostly the `createSystem` method, taken from `OpenMM` with some housekeeping around it. `OpenMM` is doing the heavy lifting there, and it is no small feat to strip that out. However, `run_atomtyping` should not require that information and should be able to run with only a `ParmEd` backend. Right now, there is a `_prepare_topology` method that takes in whatever is being passed into `.apply()` and converts it to an `OpenMM` object. To make this change, we would need to hope it's a `ParmEd` object - convert it from `mbuid.Compound` if not, and generally not take in an `OpenMM` object.

So to make this work, `run_atomtyping` and its many components needed to be updated to operate on different objects. The first thing that happens is checking if it is asked to do residue-based atomtyping; for now I am skipping over that. The driver inside `run_atomtyping` is the function `find_atomtypes` which is in `foyer/atomtyper.py`. Inside this function, the typemap is initialized, rules loaded, and some pre-processing done on those rules, and then passed on to another function `_iterate_rules`. This function iterates over the rules and atoms and stores matches in the atom's whitelist and overrides in the blacklist. So we need to go a file deeper into `foyer/smarts_graph.py`. Fortunately, almost none of the atom-typing algorithm needed to be replaced, just a long list of small changes to make it operate on a different object. Throughout this process, there were a number of 1:1 conversions (switching from `top.residues()` which looks like a generator to `structure.residues`, which is only an iterator).

There are still some things not working yet (non-element types aren't done being fixed) so tests aren't all passing, but it's coming along...

mattwthompson:
Something I can't quite figure out right now is why one of the residues is being clobbered in `parametrize_system` ....

```python3
(Pdb) structure
<Structure 11 atoms; 3 residues; 10 bonds; PBC (orthogonal); NOT parametrized>
(Pdb) topology, positions = _topology_from_parmed(structure, self.non_element_types)
(Pdb) topology
<Topology; 3 chains, 3 residues, 11 atoms, 10 bonds>
(Pdb) pmd.openmm.load_topology(topology=topology, system=system)
<Structure 11 atoms; 2 residues; 10 bonds; parametrized>

```

It happens somewhere in here:
https://github.com/mosdef-hub/foyer/blob/2c44f562b0063425ec1772292a4be3893e7a1b4e/foyer/forcefield.py#L616-L624

ahy3nz:
https://github.com/mosdef-hub/mbuild/blob/201162105900126d8cab3e7256c68cc0de820551/mbuild/compound.py#L2414
Should that `else` statement be indented one? I'm trying to figure out why there's the common error of `a.atomic_number > 0` gets you a `str` vs `int` comparison error. Even if it were a parmed custom element, I would think it's still a valid comparison?

ahy3nz:
Also, for your parmed-openmm loop, can you show your code? I didn't have that issue
[pmd_omm_loop.gz](https://github.com/mosdef-hub/foyer/files/4072470/pmd_omm_loop.gz)

mattwthompson:
>Also, for your parmed-openmm loop, can you show your code?

Sorry this was from running `py.test foyer/tests/test_forcefield.py::test_apply_residues`. Should be able to duplicate my issue (if it truly exists) with a trace anywhere in that L616 snippet

rmatsum836:
For the record, I'm getting the same behavior when I run `test_apply_residues`

rmatsum836:
This issue was caused because ParmEd residues were being indexed by `name` and `number` in a dictionary in `_topology_from_parmed`.  For whatever reason, the `number` for each residues was the same, so only 2 of the 3 residues were added to the dictionary.  

This _seems_ to be fixed if we index the residues by `name` and `idx`, not by `number`.

> Something I can't quite figure out right now is why one of the residues is being clobbered in `parametrize_system` ....
> 
> Running `py.test foyer/tests/test_forcefield.py::test_apply_residues`:
> 
> ```python
> (Pdb) structure
> <Structure 11 atoms; 3 residues; 10 bonds; PBC (orthogonal); NOT parametrized>
> (Pdb) topology, positions = _topology_from_parmed(structure, self.non_element_types)
> (Pdb) topology
> <Topology; 3 chains, 3 residues, 11 atoms, 10 bonds>
> (Pdb) pmd.openmm.load_topology(topology=topology, system=system)
> <Structure 11 atoms; 2 residues; 10 bonds; parametrized>
> ```
> 
> It happens somewhere in here:
> 
> https://github.com/mosdef-hub/foyer/blob/2c44f562b0063425ec1772292a4be3893e7a1b4e/foyer/forcefield.py#L616-L624

mattwthompson:
I am marking this as ready for review now that everything is passing (at least locally ...). Thanks to @justinGilmer @rmatsum836 @ahy3nz for helping me with some roadblocks.

Some outstanding concerns I have include 
* An MDTraj dependency is added here, just for element housekeeping.
* Some tests have been updated, but others probably need to be updated, and some behavior may not be tested well, so we may want to add more
* I'm not sure where specifically, but I get the feeling some of these functions could be trimmed down. It seems like there are a few too many steps involved in an `apply()` call.

mikemhenry:
MDTraj can be quite the pain when installing sometimes, would it be possible for us to just vendor the Element class?

ahy3nz:
It looks like we only use mdtraj's Element class to get atomic symbols from atomic numbers. Could we use parmed's periodic_table class?

https://github.com/ParmEd/ParmEd/blob/master/parmed/periodic_table.py#L45

```python3
from parmed.periodic_table import Element
atomic_symbol = Element[a.atomic_number]

system_elements.add(atomic_symbol)
```

mattwthompson:
That works perfectly, thanks Alex

foyer/atomtyper.py
```diff
@@ -34,7 +52,7 @@ def find_atomtypes(topology, forcefield, max_iter=10):
             except IndexError:
                 try:
                     atomic_num = next(atom.find_data('atomic_num')).children[0]
-                    element = pt.Element[int(atomic_num)]
+                    element = Element.getByAtomicNumber([int(atomic_num)]).symbol
```
ahy3nz: Is this statement still using mdtraj Element?

mattwthompson:
This mbuild test is failing for me when using this patch

```
    def test_save_residue_map(self, methane):
        filled = mb.fill_box(methane, n_compounds=10, box=[0, 0, 0, 4, 4, 4])
        t0 = time.time()
        filled.save('filled.mol2', forcefield_name='oplsaa', residues='Methane')
        t1 = time.time()
        foyer_kwargs = {'use_residue_map': False}
        filled.save('filled.mol2', forcefield_name='oplsaa', overwrite=True,
                    residues='Methane', foyer_kwargs=foyer_kwargs)
        t2 = time.time()
>       assert (t2 - t1) > (t1 - t0)
```

ahy3nz:
> This mbuild test is failing for me when using this patch
> 
> ```
>     def test_save_residue_map(self, methane):
>         filled = mb.fill_box(methane, n_compounds=10, box=[0, 0, 0, 4, 4, 4])
>         t0 = time.time()
>         filled.save('filled.mol2', forcefield_name='oplsaa', residues='Methane')
>         t1 = time.time()
>         foyer_kwargs = {'use_residue_map': False}
>         filled.save('filled.mol2', forcefield_name='oplsaa', overwrite=True,
>                     residues='Methane', foyer_kwargs=foyer_kwargs)
>         t2 = time.time()
> >       assert (t2 - t1) > (t1 - t0)
> ```

I wonder if it's all the overhead from making new pmdStructures for each new residue. Unwrapping the typemap and finding the atomtypes for each residue probably aren't the source of the slowdown

rmatsum836:
The `residue_map` dictionary was being constructed with `res` as the key value.  Because each `res` object is unique to each other, the `residue_map` was being incorrectly built and significantly slowed down the performance.  This caused `test_save_residue_map` in mBuild to fail.

This bug was fixed by using `res.name` as the key value in `residue_map`.

After these changes, all foyer and mbuild tests are passing locally for me.

mattwthompson:
Thanks Ray!

I checked on my PC (WSL) with a fresh mbuild clone & install and test are bassing here and in mbuild.

mattwthompson:
I think this is good to go, and the branch is up to date with master. This passes all unit tests on mbuild and mbuild-examples.

Could some of you @mosdef-hub/mosdef-contributors pull this patch and test this on your systems?

rmatsum836:
I think this looks good.

Could some of you @mosdef-hub/mosdef-contributors pull this patch and test this on your systems?

Issue #297, opened 2019-11-11
Apply multiple forcefields to a single Compound


rmatsum836:
**Describe the behavior you would like added to Foyer**
In my research I often have to apply multiple (but compatible) force fields to systems.  For the longest time, I've been doing this in my initialization scripts.  I would now like to have some function that would handle this within Foyer.

**Describe the solution you'd like**
This is a temporary solution, as I believe we can handle this situation better when our new backend is more fully developed.  However my solution right now is to iterate through the children of a compound, apply the forcefield to these individual children, and add them to an empty ParmEd structure.

A basic proof of concept here: https://gist.github.com/rmatsum836/963aeacd7936775be58b33e14e7449ce

ahy3nz:
Do you have example structure/xmls you can share? Is passing multiple forcefields via `foyer.Forcefield(forcefield_files=['a.xml', 'b.xml'])` not sufficient? Does that lead to atomtyping collisions that overrides don't cover?

rmatsum836:
I thought about this approach as well.  I think the problem there would be due to atomtyping collisions.  For example a SMARTS string of `H[C;X4]` may be sufficient to atom type a particle for a given force field, but may be too general when combined with another force field xml.

justinGilmer:
Yeah that was the problem we ran into when thinking about this a bit ago. There is no way with the current implementation to handle collisions, and overrides between two different xmls would not be handled correctly, _i think_

bc118:
In the Charmm writer for GOMC, I enabled a force field to be applied on a per residue (for GOMC this is currently the same as a molecule).  It sorts and creates N bins for N residues and force fields them as a group with their respective force fields.   This fixes the issues with errors from over lapping force fields and having to modify a large force field xml file to change water model , for example.

daico007:
Now handled by the new `https://github.com/mosdef-hub/gmso/tree/develop` and https://github.com/mosdef-hub/forcefield-utilities`

Issue #262, opened 2019-08-30
Should we drop support for loading in multiple XMLs?


mattwthompson:
**A summary of the question or discussion topic.**

Right now you can instantiate a force field object with multiple files (`Forcefield(forcefield_files=['ff1.xml', 'ff2.xml'])`) which, as best I can understand it, is a holdover from how OpenMM assumes you have different XMLs for protein and water.

* I don't think we have a smart way of handling precedence (i.e. if water atom types are defined in each XML file, which should be used?)

* I don't think anybody actually uses this feature?

Therefore I think we should consider dropping it. Most of the code would remain the same, it would just be loading, processing, etc. one file instead of a list-like.

Issue #153, opened 2017-12-18
Forcefield.name attribute?


mattwthompson:
In at least one of the atom typing errors, it is hard-coded to report that it found something wrong with the OPLS types whether or not we are using OPLS. For example:

```
AssertionError: Found inconsistent OPLS types in Generic title: [(10, 'mycoolff_044', 'mycoolff_045')]
```

I could simply change this to report `blah blah atom types ...` but maybe better would be to have it report the force field name. This would, however require adding a `name` attribute to the `Forcefield` class. I was thinking it would be optional, and would only take up one line of an xml. This should only require small changes.

Adding an attribute for a marginally cooler error message seems like overkill, but it seems natural to me that it should have one. We already can initiallize a `Forcefield` using a `name` argument (though this means something slightly different):

```
>>> OPLSAA = Forcefield(name='oplsaa')
KeyboardInterrupt
>>> OPLSAA.name
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Forcefield' object has no attribute 'name'

```

Thoughts?

ctk3b:
Good catch. I think a first really simple PR will be to simply strike out the OPLS references [here](https://github.com/mosdef-hub/foyer/blob/b5bf96356c40069f110a907a749be6fb6ed62415/foyer/tests/utils.py#L30-L48). 

I like the name suggestion also but maybe let's do that in a separate PR since that requires deciding precisely how to name forcefields. Is simply going to be helpful metadata or something that we want to rely on to be precise? Does the .xml file name encode the forcefield name or do we add a separate xml tag?

mattwthompson:
> Good catch. I think a first really simple PR will be to simply strike out the OPLS references here.

👍 Exactly what I was about to do.

> I like the name suggestion also but maybe let's do that in a separate PR since that requires deciding precisely how to name forcefields. Is simply going to be helpful metadata or something that we want to rely on to be precise?

My thinking was making this an optional argument and attribute would help us avoid some of these harder issues.

> Does the .xml file name encode the forcefield name or do we add a separate xml tag?

Except this as the immediately confusing one. I think the metadata in the xml should be considered more important, but we should also think about better ways to access force field files than their name. I'm starting to use multiple force fields simultaneously and while it's working well enough so far, accessing them by file name seems clunky to me.

summeraz:
[...]

Issue #149, opened 2017-11-21
Apply multiple force fields simultaneously


mattwthompson:
Often when dealing with interesting physical systems, there are multiple heterogeneous components and it is necessary to apply different force fields to portions of a system. I've done this a couple times, but always in a hacky way

1. Separate the input structure. Make two or more compounds from one by cloning children into new compounds) and apply separate force fields to each. Then, add together in parmed.

2. Combine the force field file by hand and apply it to the whole system. Only feasible for simple cases with small force fields and no common atom types, such as a system of water on graphene.

Ideally we could do this in one pass, maybe based off of residue name, like

```
aotmtyped_structure = foyer.apply_multiple(forcefield=[TIP3P, AMBER], resnames=['SOL', 'METH'])
```

There are plenty of tricky details to figure out, such as deciding which force field has priority, and probably a handful more that we would encounter down the road.

I think we've discussed this before, but I can't find an issue. This sort of thing would be a nice feature, but might also require a rework on the scale of a 1.0 release.

davidlmobley:
We point out to users in `openforcefield` that approach (1) is simple, and give an example highlighting how to apply a conventional AMBER forcefield for a protein in combination with our SMIRNOFF force field for a small molecule. As you point out, with ParmEd this is straightforward.

PR #298, opened 2019-11-22
Add version attribute to force fields

### LINKED ISSUE #187: Maintaining different versions of force fields

mattwthompson:
### PR Summary:

Fixes  #187, would be useful in  #290 

* Adds `version` as an attribute of the `ForceField` element in the XML schema
* Adds `version` as an attribute to the `Forcefield` class
* Adds a simple test via reading a toy XML file and asserting the above two `version`s are equal

This is not done elegantly right now; 99% of the XML loading is done by OpenMM, so I had to do this after the fact and only dealing with the case of passing a single XML, which is all I believe we do right now. I'm not sure what the target for this PR should be, but I think we should soon have a bigger discussion about how we want to handle multiple XML files. The OpenMM use case is basically `Forcefield('protein_model.xml', 'water_model.xml')`, i.e. there is a clean interface between constituents and not any expected overlap between the two. This doesn't cleanly map on to our use cases, I don't think. Related to #297 #262 #149

If this seems reasonable I can do the same for names (#153), here or in another PR. Requesting a few specific reviewers since I have discussed this with many people over the past year or two (but only gotten to it now 🤕)

### PR Checklist
------------

### DIFF
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -343,6 +343,7 @@ class Forcefield(app.ForceField):
         self._included_forcefields = dict()
         self.non_element_types = dict()
         self._version = None
+        self._name = None
 
         all_files_to_load = []
         if forcefield_files is not None:
@@ -372,8 +373,10 @@ class Forcefield(app.ForceField):
 
         if isinstance(forcefield_files, str):
             self._version = self._parse_version_number(forcefield_files)
+            self._name = self._parse_name(forcefield_files)
         elif isinstance(forcefield_files, list):
             self._version = [self._parse_version_number(f) for f in forcefield_files]
+            self._name = [self._parse_name(f) for f in forcefield_files]
 
         self.parser = smarts.SMARTS(self.non_element_types)
         self._SystemData = self._SystemData()
@@ -381,6 +384,10 @@ class Forcefield(app.ForceField):
     @property
     def version(self):
         return self._version
+
+    @property
+    def name(self):
+        return self._name
 
     @property
     def included_forcefields(self):
@@ -402,6 +409,15 @@ class Forcefield(app.ForceField):
             root = tree.getroot()
             try:
                 return root.attrib['version']
+            except KeyError:
+                return None
+
+    def _parse_name(self, forcefield_file):
+        with open(forcefield_file, 'r') as f:
+            tree = ET.parse(f)
+            root = tree.getroot()
+            try:
+                return root.attrib['name']
             except KeyError:
                 return None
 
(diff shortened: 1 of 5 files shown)


ahy3nz:
```
 /Users/ayang41/Programs/foyer/foyer/tests/test_forcefield.py(511)test_load_version_number()
-> assert lj_ff.version == '0.4.1'
(Pdb) l
506  	            assert str(item.xpath('comment()')) == '[<!--Note: original overrides="opls_144"-->]'
507
508  	def test_load_version_number():
509  	    lj_ff = Forcefield(get_fn('lj.xml'))
510  	    import pdb; pdb.set_trace()
511  ->	    assert lj_ff.version == '0.4.1'
[EOF]
(Pdb) foo = Forcefield(forcefield_files=[get_fn('lj.xml')])
(Pdb) foo.version
'foobar'
(Pdb) lj_ff.version
'0.4.1'
(Pdb)
```

* You have an extra `foobar` debug line in the init
* Did you want to cover the case where someone passes a list of XML files? This PR appears to only cover the case of passing a single XML string, not a list (which is what most people tend to do anyways...)

mattwthompson:
> Did you want to cover the case where someone passes a list of XML files? This PR appears to only cover the case of passing a single XML string, not a list (which is what most people tend to do anyways...)

I should be able to rework it to handle multiple files but I also think this might be a feature we want to drop at some point.

mattwthompson:
I'm happy with where this is at now. If this approach is ok, I will duplicate this functionality to support a `name` attribute.

foyer/forcefield.py
```diff
@@ -385,6 +403,24 @@ def included_forcefields(self):
             self._included_forcefields[basename] = ff_filepath
         return self._included_forcefields
 
+    def _parse_version_number(self, forcefield_file):
+        with open(forcefield_file, 'r') as f:
+            tree = ET.parse(f)
+            root = tree.getroot()
+            try:
+                return root.attrib['version']
+            except KeyError:
```
justinGilmer: Going off of Mike's comment about raising a warning, this seems like the place where that would go.
```suggestion
            except KeyError:
                warn.Warn("version warning message here")
                return None
```

PR #300, opened 2019-11-25
Fix Travis Build Erros

umesh-timalsina:
This is a very minor change to download `Minconda` using `curl` rather than `wget`. The errrored builds we are getting is due to `wget` and `openssl` paths. Using `curl` resolves this.

Travis build for this branch can be tracked [here](https://travis-ci.org/umesh-timalsina/foyer/jobs/616827616).


mattwthompson:
I've always used `wget` and `curl` interchangeably, I don't know if there's any downside to one or the other.

Also worth noting we'll want to make this same change to mBuild and any other repos for the time being. However, as we've discussed in other channels, I think we can either drop macOS builds from travis and/or drop travis altogethers

mattwthompson:
Thanks @umesh-timalsina!

PR #302, opened 2019-12-26
Don't install md5sha1sum

mattwthompson:
### PR Summary:

Same as https://github.com/mosdef-hub/mbuild/pull/660

PR #303, opened 2019-12-26
Refactor tests using mbuild examples to using mbuild recipes

ahy3nz:
### PR Summary:
Should help #302 
### PR Checklist
------------

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ env:
 
 install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install md5sha1sum; fi
   - source devtools/travis-ci/install_conda.sh
   - conda config --set always_yes yes --set changeps1 no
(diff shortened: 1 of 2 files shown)


ahy3nz:
It seems like this only works for bleeding mbuild (where we removed `mbuild.examples`), but for the other mbuild releases, `mbuild.recipes` isn't a valid module?

ahy3nz:
So here's the problem:
29 days ago, when mbuild 0.10.3 was released, `mbuild.recipes.Alkane` was not a valid module, as the `setup.py` entrypoints had not been updated. 24 days ago, `mbuild.recipes.Alkane` DID become a valid module via the entrypoints. So only the bleeding-mbuild tests are passing due to `import mbuild.recipes.Alkane`, whereas all the conda-mbuild tests are failing because of that same import

Solution: Do another mbuild release? Keep refactoring the foyer tests to avoid using any semblance of an `Alkane` classs?

mattwthompson:
I think a release here is necessary. If we want that to be a valid module, and I think we do, it should be reflected on Anaconda.

PR #312, opened 2020-02-06
Try building on Python 3.8 using conda-forge/label/testing

mattwthompson:
### PR Summary:

Mirroring https://github.com/mosdef-hub/mbuild/pull/650

### PR Checklist
------------

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,10 @@ matrix:
   include:
     - { os: linux, env: PYTHON_VERSION=3.6 }
     - { os: linux, env: PYTHON_VERSION=3.7 }
+    - { os: linux, env: PYTHON_VERSION=3.8 }
     - { os: osx, env: PYTHON_VERSION=3.6 }
     - { os: osx, env: PYTHON_VERSION=3.7 }
+    - { os: osx, env: PYTHON_VERSION=3.8 }
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


justinGilmer:
What is the status of this currently? I am a bit behind on this one.

mattwthompson:
The status is basically that I haven't put in the time and effort to get the CI bots to pull from the right label, and I'm not sure it's worth figuring out now vs. waiting for OpenMM to properly be on forge in some weeks or months' time

mattwthompson:
Just a few updates:

* Moving OpenMM over to `conda-forge` is (evidently) taking longer than hoped as of a few months ago, and for now it would be best to treat is as an omnia-only package.
* Other issues are preventing OpenMM from being built on Python 3.8. I don't know how long that will take to be resolved, but they are likely independent from the forge migration.
* OpenMM is the only dependency holding foyer back from being built on Python 3.8. Commenting it out from `requirements-dev.txt` makes things install just fine (`conda create --name foyer-py38 --file requirements-dev.txt python=3.8 -c conda-forge -c omnia -c mosdef` pulls in 3.8 packages on my Linux machine)
* Given that the CUDA parts of OpenMM are not needed here, it may be possible to monkey-build a special "I basically only need the `Forcefield` class" build of OpenMM and host it on `mosdef`, although the downsides are obvious so I would not generally recommend that.

In summary, stripping out the OpenMM backend from foyer is probably the quickest path to building the MoSDeF ecosystem on Python 3.8. This may sound like a heroic task but I've done a few things to set the stage for it (#296 in particular) and much of the functionality should end up in GMSO anyway. Moving on from Python 3.6 may sound premature, but it's already on security-only releases and will be EOL at some point in 2021 (unless `conda-forge` stops build it, in which case it's effectively EOL then).

mattwthompson:
If somebody wanted to pick up a similar idea ("I have a package with a known list of deps and want to see if I can install Python x.y") you can modify this script that uses the conda API: https://github.com/mattwthompson/scripts/blob/789be67585e56746bb58929f6fe3ed8da90ce3a2/check-py38.py

Issue #326, opened 2020-03-09
Error when `verbose=True` for Periodic Dihedrals


rmatsum836:
**Bug summary**

This was reported back in #323 but currently `verbose=True` fails when a system with only periodic dihedrals is atom-typed.  The reason for this is that `verbose` only checks for rb torsions.  This code attempts to love over `structure.rb_torsions` which has a length of 0, so the `pmd_ids` variable never gets set.  Thus the error occurs when the print statement attempts to call` pmd_ids`.

I don't think this issue is going to be fully solved as long as ParmEd is our backend, but I think we should at least prevent the atom typing step from failing if `verbose==True`

**Code to reproduce the behavior**

Error message: ```UnboundLocalError: local variable 'pmd_ids' referenced before assignment```

**Possible solutions**

1. Continue only looping through structure.rb_torsions, but have a try, except statement when printing out missing propers

```
try:
    print('missing proper with ids {}'.format(pmd_ids))
except UnboundLocalError:
    print('No RB Torsions found')
```

2. Loop through BOTH `structure.rb_torsions` and proper dihedrals.  This is still kind of hacky because of the layered nature of dihedrals (In this case there are more parametrized dihedrals than system dihedrals).  As a result, a message wouldn't be printed in this case.

chrisiacovella:
I just ran into this issue, as well.  Also the code as it is doesn't give any useful information if you aren't using RB torsions.  I'll submit a PR, but this change seems to fix it for both periodic and rb torsions:

```
    if verbose:
        for omm_ids in data.propers:
            missing_dihedral = True
            for pmd_proper in proper_dihedrals:
                pmd_ids = (pmd_proper.atom1.idx, pmd_proper.atom2.idx, pmd_proper.atom3.idx, pmd_proper.atom4.idx)
                if pmd_ids == omm_ids:
                    missing_dihedral = False
            for pmd_proper in structure.rb_torsions:
                pmd_ids = (pmd_proper.atom1.idx, pmd_proper.atom2.idx, pmd_proper.atom3.idx, pmd_proper.atom4.idx)
                if pmd_ids == omm_ids:
                    missing_dihedral = False
            if missing_dihedral:
                print("Missing dihedral with ids {} and types {}.".format(
                        omm_ids, [structure.atoms[idx].type for idx in omm_ids]))
```

Issue #323, opened 2020-03-02
Foyer should identify missing parameters when it fails


jpotoff:
When Foyer fails to assign parameters especially for dihedrals), it should identify the parameter types that it could not find in the XML file. 

[example.zip](https://github.com/mosdef-hub/foyer/files/4276833/example.zip)

jpotoff:
Replaced the vapor_box_param assignment with this, but did not get particularly useful output.
vapor_box_param=ff.apply(vapor_box,verbose='True')

UnboundLocalError                         Traceback (most recent call last)
<ipython-input-33-5a25f00916e3> in <module>
      2 
      3 ff=foyer.Forcefield(forcefield_files='opls_gomc.xml')
----> 4 vapor_box_param=ff.apply(vapor_box,verbose='True')
      5 #liquid_box_param=ff.apply(liquid_box)
      6 par_writer.write_par(vapor_box_param,'mbuild.par')

~/anaconda3/lib/python3.7/site-packages/foyer/forcefield.py in apply(self, topology, references_file, use_residue_map, assert_bond_params, assert_angle_params, assert_dihedral_params, assert_improper_params, combining_rule, verbose, *args, **kwargs)
    499             assert_angle_params=assert_angle_params, assert_dihedral_params=assert_dihedral_params,
    500             assert_improper_params=assert_improper_params, combining_rule=combining_rule,
--> 501             verbose=verbose, *args, **kwargs)
    502 
    503     def run_atomtyping(self, topology, use_residue_map=True):

~/anaconda3/lib/python3.7/site-packages/foyer/forcefield.py in parametrize_system(self, topology, positions, references_file, assert_bond_params, assert_angle_params, assert_dihedral_params, assert_improper_params, combining_rule, verbose, *args, **kwargs)
    567         _check_angles(data, structure, verbose, assert_angle_params)
    568         _check_dihedrals(data, structure, verbose,
--> 569                               assert_dihedral_params, assert_improper_params)
    570 
    571         if references_file:

~/anaconda3/lib/python3.7/site-packages/foyer/forcefield.py in _check_dihedrals(data, structure, verbose, assert_dihedral_params, assert_improper_params)
    297                     missing_dihedral = False
    298             if missing_dihedral:
--> 299                 print('missing improper with ids {}'.format(pmd_ids))
    300 
    301     if data.propers and len(data.propers) != \

UnboundLocalError: local variable 'pmd_ids' referenced before assignment

[...]

PR #327, opened 2020-03-10
Add temporary fix for periodic dihedrals

rmatsum836:
### PR Summary:
Addressing issues raised in #323 and #326.  Two changes in `forcefield.py`:
- when `verbose=True`, compare the openmm dihedrals to both parmed rb_torsions and dihedrals.  This allows a system with periodic dihedrals to be compared with `structure.dihedrals`.
- Raise separate errors/warnings for when openmm dihedrals != parmed dihedrals vs openmm dihedrals < parmed dihedrals.  The motivation for this is in the case of #323, there were more parmed dihedrals than openmm dihedrals because a single dihedral type contained multiple dihedrals.  I don't think this should raise an error, so in this case I raise a warning and a separate message instead.

This is very much a temporary fix, and I expect dihedrals will be handled much better with the new backend.  This is also just an idea, so feel free to tell me if it's a bad one.

### PR Checklist
------------

### DIFF
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -310,22 +310,32 @@ def _check_dihedrals(data, structure, verbose,
     if verbose:
         for omm_ids in data.propers:
             missing_dihedral = True
-            for pmd_proper in structure.rb_torsions:
+            for pmd_proper in (structure.rb_torsions + proper_dihedrals):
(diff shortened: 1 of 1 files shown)


daico007:
Closing this PR since the issue has been fixed by #335

Issue #336, opened 2020-06-09
Warning for non-zero net charge


rsdefever:
**Describe the behavior you would like added to Foyer**
Add a warning if the net charge of the parameterized system is non-zero. The motivation is that in many cases a non-zero net charge indicates that the atom typing did not work correctly. It is not a perfect solution as there are obviously cases where the molecule should have a net charge, e.g., atom typing a IL ion, neutral systems, luck, etc. However, it will help catch some potential problems. See **additional context** for more detail.
**Additional context**
This idea was motivated by an experience I had; I used foyer to atom type a naphthalene molecule. Everything ran fine, no warnings or anything. it was only after I saved the topology file to disk that I realized the fusion C atom type for naphthalene does not exit in the foyer OPLS-AA FF XML file yet. A warning that my molecule had a non-zero charge post atom typing would have been an earlier indicator that something had gone awry.

rmatsum836:
This is an issue I run into occasionally as well and I think that's a good idea.  I think it would be a fairly easy check to implement at the end of `parametrize_system`.  I'm also wondering if we want to have a net_charge argument that is `0.0` by default.  That way if a user is atom typing systems with a non-zero charge, the user will also be notified if the charge of their system doesn't match the desired net charge.

PR #338, opened 2020-06-11
Add warning for non-zero charge structure

daico007:
Relate to #336. Add a warning if a typed structure has non-zero charge. 'np.allclose()' is used instead of '==' to check for 0 equivalence so since some system may have insignificant non-zero net charge (like 1e-16).

### PR Checklist
------------


rsdefever:
Looks great. Can you add a couple of simple unit tests and then LGTM!

PR #345, opened 2020-06-23
Entrypoints conda

justinGilmer:
### PR Summary:
To ensure that we have our plugins supported properly when installed
from conda, it appears that a few options are necessary to ensure that
is the case.

For the case of setuptools entry_points, the conda-build documentation
recommends that we do include the build option: preserve_egg_dir

https://docs.conda.io/projects/conda-build/en/latest/user-guide/recipes/build-without-recipe.html#preserve-egg-directory

Setuptools are also now build and runtime dependencies.

### PR Checklist
------------

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -11,13 +11,11 @@ build:
     - python
   number: {{ environ.get('GIT_DESCRIBE_NUMBER', 0)}}
   script: python setup.py install --single-version-externally-managed --record record.txt
-  entry_points:
-          - load_OPLSAA = foyer.forcefields.forcefields:load_OPLSAA
-          - load_TRAPPE_UA = foyer.forcefields.forcefields:load_TRAPPE_UA
 
 requirements:
   build:
     - python
+    - setuptools
 
   run:
     - parmed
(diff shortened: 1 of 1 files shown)


daico007:
Yeah, I think there is one way to test this PR: pull this PR and build it locally, then create a new environment and conda install that local package. It's quite a clumsy method but I think it will show if this PR do what we expect.

justinGilmer:
I was thinking about that situation a lot last night and this morning, and I am not sure if there is an easy way to test this. I think we would need one of the plugins to be a conda install-able package with the same changes above in its own `meta.yml`. but I did conda-build it locally last night, and installed an editable version of [gaff_foyer](https://github.com/rsdefever/GAFF-foyer) with pip and that worked no problem.

rmatsum836:
I just tested this and it's working for me.  For future reference what was the steps you took?  Here is what I did:

1. create separate conda environment
2. git clone foyer, add your fork as a remote and checkout into your branch
3. Run `conda build meta.yml`
4. Run `conda install --use-local path_to_foyer.bz`
5. install other required dependencies
6. Run `pip install -e .` in GAFF-foyer
7. Run `python -c "import python; foyer.forcefields.load_GAFF()"`

PR #348, opened 2020-07-14
GMSO integration 

daico007:
### PR Summary:
The final goal of this PR is to add GMSO as another backend for foyer. This is still a WIP, more updates will be posted in the comment as I progress. 

### PR Checklist
------------

### DIFF
--- a/foyer/atomtyper.py
+++ b/foyer/atomtyper.py
@@ -6,7 +6,6 @@ import parmed.periodic_table as pt
 from foyer.exceptions import FoyerError
 from foyer.smarts_graph import SMARTSGraph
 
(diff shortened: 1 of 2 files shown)


daico007:
Foyer atomtyper is working for GMSO Topology now (tested with an Ethane, but haven't tried anything fancy yet)! 
So after some discussion with folks at Vandy, I decided to add a `'topology'` key in `typemap` to store the Topology object, this seems to solve our issue with not being able to get the index and neighbors count for any Atom/Site.
I will continue to test and see if this works fine with more complicated system, and continue working on the parameterizing step.

PR #349, opened 2020-07-24
Boil down graph

ahy3nz:
Looking at https://github.com/mosdef-hub/foyer/pull/348, I'm concerned there's a lot of spaghetti code to handle both parmed and gmso objects. I think for some "upstream" foyer processing, it might be an important stopgap to accommodate both objects, but a lot of the downstream graph matching code within foyer's `smarts_graph.py` shouldn't have to change. Namely, generalize the properties of the topological graph's nodes to be more "primitive" properties (like index, symbol, number, n_bond_partners) instead of the node's singular property being a bulky parmed atom (or gmso site) which contains the relevant properties. This PR tries to address this:

* Rework `smarts_graph.find_matches` so the properties of the topological graph are the primitive properties of each atom, not a parmed atom that wraps the primitive properties. This should make graph processing slightly more lightweight so that the parmed objects don't have to be carried around. This also establishes a pattern for swapping out parmed objects with gmso objects.

* Refactor `smarts_graph._atom_id_matches` to accommodate the new design of the topological graph. We no longer need to access atomic properties of the parmed atoms via the properties of the networkx node; we only need to access atomic properties stored directly within the networkx node

This was putting code to paper to demonstrate the idea/solution. But my primary concern still remains: when making a stable transition from parmed to gmso, we should try to minimize the amount of duplicate code. This could be done by trying to generalize downstream code to be less library-dependent, but upstream processing code may still have to be library-dependent

### DIFF
--- a/foyer/atomtyper.py
+++ b/foyer/atomtyper.py
@@ -154,40 +154,22 @@ def _iterate_rules(rules, topology, typemap, max_iter):
 
     """
 
-    if isinstance(topology, parmed.Structure):
-        for _ in range(max_iter):
-            max_iter -= 1
-            found_something = False
-            for rule in rules.values():
-                for match_index in rule.find_matches(topology, typemap):
-                    atom = typemap[match_index]
-                    # This conditional is not strictly necessary, but it prevents
-                    # redundant set addition on later iterations
-                    if rule.name not in atom['whitelist']:
-                        atom['whitelist'].add(rule.name)
-                        atom['blacklist'] |= rule.overrides
-                        found_something = True
-            if not found_something:
-                break
-        else:
-            warn("Reached maximum iterations. Something probably went wrong.")
-    elif isinstance(topology, gmso.Topology):
-        for _ in range(max_iter):
-            max_iter -= 1
-            found_something = False
-            for rule in rules.values():
-                for match_index in rule.find_matches(topology, typemap):
-                    atom = typemap[match_index]
-                    # This conditional is not strictly necessary, but it prevents
-                    # redundant set addition on later iterations
-                    if rule.name not in atom['whitelist']:
-                        atom['whitelist'].add(rule.name)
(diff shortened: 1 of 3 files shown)


ahy3nz:
If this transition-design has already been discussed, my bad

daico007:
This is great! Yeah my code for #348 is quite messy now, probably it gonna get worse when I try to switch foyer inheritance from openmm forcefield to gmso forcefield (I am planning to use multi inheritance with a keyword switch for the transition stage, please let me know if you have any recommendation). Nonetheless, simplify and generalize the downstream code would definitely help simplify a lot of stuff!

daico007:
so for the current approach, my thought is to have a check at the very beginning of any method that have parmed specific syntax (`if isinstance` stuff), and my plan is that once the gmso integration is stable, we can just delete the parmed/openmm block. Yeah, but simplify the code now could help a lot with not inflating our code base.

daico007:
And one tiny thing, @ahy3nz can you change so this PR will merge to the `gmso_integration` branch of foyer?

daico007:
I just changed so this PR will merge into `foyer:gmso_integration`, so this can move a bit quicker.

ahy3nz:
Okay I touched up atomtyper.py, smarts_graph.py, and test_graph.py. I think this is all I wanted to accomplish for those two modules.  Let me know if the changes make sense. 

Hopefully this adds some refactoring and code-simplifying. Some foyer processing logic changed a little to operate on more-generic networkx graphs and avoid passing around library-dependent objects. Tests were passing locally but I don't expect them to pass here because of the gmso imports

When you get around to forcefield.py I'll try to pop in again

daico007:
Cool, thank you! I will look at this today or tomorrow and merge it in (since this is being merged to a feature branch, I don't think I am going to wait for 2 approvals). I will ping you when I push the code for forcefield.py!

PR #355, opened 2020-08-04
Manage pypi  and docker publish using azp (Closes #354)

### LINKED ISSUE #354: Manange pypi releases with azure pipelines
We should manage `pypi` releases with azure pipelines. Similar to https://github.com/mosdef-hub/mbuild/issues/665. 
We should also manage docker images push.

umesh-timalsina:
Closes #354.


mikemhenry:
Why do we publish on this package on pypi at all and not just exclusively publish it on conda-forge? Since the dependencies for `mbuild` require some software that isn't on pypi, I worry that users will `pip install` it and then just face import errors. Sorry for only bringing this up now but as I was thinking about the migration from `pypi test` to `pypi` I realized that there are some issues with it being on `pypi`.

umesh-timalsina:
I agree. But, we have been publishing to `pypi` (manually). This PR was trying to automate that.

daico007 (approved):

PR #358, opened 2020-08-25
GMSO integration

daico007:
### PR Summary:
Continue the GMSO integration process. This time is to create a new foyer.Forcefield class inherit from the GMSO Forcefield object instead of from OpenMM Forcefield. 

Initial plan includes rewrite the `from_xml` method to overwrite the one in GMSO forcefield. The new `from_xml` method will create the Forcefield object from foyer XML and parse other foyer specific information. Then we can add other methods necessary to do the atomtyping step. 

### PR Checklist
------------


ahy3nz:
So there are now 3 force field objects within foyer? One that inherits from the openmm force field (and this one can do all foyer atom-typing/parametrizing), one that inherits from the gmso force field (and this one can do all foyer atom-typing/parametrizing), and a third that inherits from nothing. This third one is a little curious, so it contains the gmso FF or the openmm FF depending on how you instantiate it. Is this goal for a more generic wrapper so there's a standard API for foyer to operate on both the foyer-gmso FF and foyer-omm FF objects? Should this general_FF contain the foyer-gmso FF or foyer-omm FF instead of the gmso FF or omm FF? This way, the general_FF object can perform any relevant foyer atom-typing/parametrization functions

daico007:
@ahy3nz  I deleted that gmos_ff.py locally, but might have forgotten to push it. The main object from this PR would be that general_forcefield object that can take use both GMSO.ForceField and OpenMM.Forcefield (to store the AtomType, BondType, etc. info). The atomtyping step worked out of the box and I am working on the parametrizing step now. I just left the old foyer Forcefield object that inherit intact for now just in case we need something for comparison (like if the new object is missing any important features).

foyer/general_forcefield.py
```diff
@@ -0,0 +1,417 @@
...
+            for ff_file_name in preprocessed_files:
+                Validator(ff_file_name, debug)
+
+        # Load in an internal forcefield object depends on given backend
+        if backend == 'gmso':
+            self._parse_gmso(*preprocessed_files)
+        elif backend == 'openmm':
+            self._parse_mm(*preprocessed_files)
+        elif backend == 'openff':
+            self._parse_ff(*preprocessed_files)
```
ahy3nz, line 149: Should these be `self._parse_gmso(preprocessed_files)` without the asterisk? These functions appear to accept a singular list, not the unwrapped, arbitrary number of arguments

foyer/general_forcefield.py
```diff
@@ -0,0 +1,417 @@
...
+            member_types = [atom.atom_type.name for atom in dihedral.connection_members]
+            dtype_name = '~'.join(member_types)
+            dihedral.dihedral_type = self.ff.dihedral_types.get(dtype_name)
+
+        # Assing ImproperTypes
+        for improper in top.impropers:
+            # Still need to deal with permutation
+            member_types = [atom.atom_type.name for atom in improper.connection_members]
+            itype_name = '~'.join(member_types)
+            improper.improper_type = self.ff.improper_types.get(itype_name)
```
ahy3nz, line 341: It looks like this implicitly assumes `self.ff` is a gmso FF? What if it's an omm FF?

If there ends up having to be a lot of very-similar-but-not-exactly-the-same code, I wonder if it would be cleaner to create separate classes, or just separate functions, to handle the different FF backends.

foyer/general_forcefield.py
```diff
@@ -0,0 +1,417 @@
...
+            Spped up options for duplicates subtopology.
+        """
+
+        if use_residue_map:
+            # Detect duplicates subtopology/residues
+            # (do matching by name, assert same number
+            # of atoms)
+            typemap = find_atomtypes(top, forcefield=self)
+        else:
+            typemap = find_atomtypes(top, forcefield=self)
```
ahy3nz, line 282: I know we used dictionaries as residue maps before, but I'm wondering if you could use [functools LRU caching](https://docs.python.org/3/library/functools.html), where the inputs to the function are some molecule's networkx graph, and the LRU caching means sometimes the function doesn't need to re-run if it's already cached the results from a previous function all

foyer/general_forcefield.py
```diff
@@ -0,0 +1,417 @@
...
+        """Parse meta data information when using OpenFF as backend
+        """
+        self.ff = app.ForceField(forcefield.files)
+        return None
+
+
+    def apply(self, top, references_file=None, use_residue_map=True,
+    assert_bond_params=True, assert_angle_params=True,
+    assert_dihedral_params=True, assert_improper_params=True,
+    verbose=False, *args, **kwargs):
```
ahy3nz, line 224: Indent

foyer/general_forcefield.py
daico007: I will fix that

foyer/general_forcefield.py
daico007: the `self.ff` can store both gmso FF and openmm FF. i was planning to handle the slight difference between gmso FF and mm FF by an `if else` block, but having a separate function is another good option too

foyer/general_forcefield.py
daico007: Thank you! I will look into that

foyer/atomtyper.py
```diff
@@ -66,7 +68,7 @@ def find_atomtypes(structure, forcefield, max_iter=10):
             except IndexError:
                 try:
                     atomic_num = next(atom.find_data('atomic_num')).children[0]
-                    element = Element[atomic_num]
+                    element = pt.Element[atomic_num]
```
bc118, line 71: A question/comment.  I'm not saying this needs to be addressed in this PR, just something that came to mind while looking at this section of code.  

In the future, are we going to make a gmso function to get the elements so the gmso topology is stand-alone from the parmed structure?

daico007: yes, that's our intention. This issue is addressed in the other PR https://github.com/mosdef-hub/foyer/pull/395

daico007:
I just looked azure error and the tests are failing because the new features of `GMSO` used in this PR (`tag` system and wildcard look up, etc) are not in a released version.

foyer/general_forcefield.py
```diff
@@ -0,0 +1,539 @@
...
+                          gmso_xml=str(tempfile.name), overwrite=True)
+            tmp_processed_files.append(tempfile.name)
+
+    else:
+        raise FoyerError('Backend not supported')
+
+    return tmp_processed_files
+
+
+class Forcefield(object):
```
bc118, line 94: This is a general comment to think about for the development process.  

When Umesh was reviewing my GOMC PR, the forcefield_files and name were in a separate selection.  In the GOMC case, the FFs are set as a single string (covers all molecules) or by a specific residue selection (which is molecule selection) via a dictionary ({Residue_name: forcefield_file or name}).  

This approach combines the forcefield_files and name into 1 variable and allows the selection of an existing/standard foyer package and selecting the many forcefield_files.  This method ensures or at least helps minimize overlapping molecule types between the selected FFs, which may be difficult for some new users.

I'm not sure it would work directly without changing a lot of things.  However, for me and the GOMC code, it did simplify it quite a bit.  This was circumvented by pre-processing the molecules into residue/molecule bins before the FF process.  In the GOMC case, we can use this dictionary to utilize both the standard foyer FF and add other xml FF files.  For example: 
 -The selection based on residues/molecule type also allows the user to add a different water FF than the existing one (say TIP3P instead of TIP4P for the oplsaa FF), which simplifies the user input to creating a TIP3P xml, rather than messing with the oplsaa FF and risk messing it up. 

The GOMC case likely will be covered in the way listed above either way, but it is just a general comment.  Again, not saying it needs to be fixed in this PR.

foyer/general_forcefield.py
```diff
@@ -0,0 +1,539 @@
...
+              assert_dihedral_params=True, assert_improper_params=True,
+              combining_rule='geometric', verbose=False, backend='gmso',
+              *args, **kwargs):
+        """Apply the force field to a molecular topology
+
+        Parameters
+        ----------
+        top : gmso.Topology, or mb.Compound
+            Molecular Topology to apply the force field to
+        references_file : str, optional, defaut=None
```
bc118, line 213: refrences misspelled -> references

foyer/general_forcefield.py
```diff
@@ -0,0 +1,539 @@
...
+            If True, Foyer will exit if parameters are not found for all
+            system angles.
+        assert_dihedral_params : bool, optional, default=True
+            If True, Foyer will exit if parameters are not found for all
+            system dihedrals.
+        assert_improper_params : bool, optional, default=True
+            If True, Foyer will exit if parametes are not found for all
+            system impropers.
+        combining_rule : str, optional, default='geometric'
+            The combining rule of the system, stored as an attribute of the
```
bc118, line 233: The default combining_rule is listed.  Maybe show the other options name (Example LB.... )

foyer/general_forcefield.py
```diff
@@ -0,0 +1,539 @@
...
+            GMSO topology.
+        verbose : bool, optional, default=False
+            If True, Foyer will print debug-level information about notable or
+            potential problematic details it encounters.
+        backend : str, optional, default='gmso'
```
bc118, line 238: The default backend value/string is listed. Maybe show the other options in the dockstring

foyer/general_forcefield.py
```diff
@@ -0,0 +1,539 @@
...
+            backend=backend,
+            combining_rule=combining_rule,
+            *args, **kwargs)
+
+    def _run_atomtyping(self, top, use_residue_map=True, **kwargs):
+        """Atomtype the topology
+
+        Parameters
+        ----------
+        top : gmso.Topology
```
bc118, line 274: "Molecular Topology to find atom types for. " sounds a little weird to me.  Maybe "Molecular Topology which is utilized to determine atom types?

In "Spped up options for duplicates subtopology" , Speed is spelled wrong.  
-Also a suggestion (not required), it may be more informative for new users to list "A speed-up option that utilizes previously atom typed molecules as a template for future atom typing of identical molecules, instead of reevaluating them each individually."

foyer/general_forcefield.py
```diff
@@ -0,0 +1,539 @@
...
+            Molecular Topology to find atom types for.
+        use_residue_map : bool, optional, default=True
+            Spped up options for duplicates subtopology.
+        """
+
+        if isinstance(top, mb.Compound):
+            top = from_mbuild(top)
+        # TO DO in another PR
+        if use_residue_map:
```
bc118, line 283: Maybe I just don't understand what is going on here.  However, it seems that it is doing the same function regardless of whether use_residue_map is True or False.

foyer/general_forcefield.py
```diff
@@ -0,0 +1,539 @@
...
+                        combining_rule='geometric',
+                        debug=False,
+                        **kwargs):
+        """Parametrize the Topology from the typemap provided
+
+        Assign AtomTypes and BondTypes to Atoms and Bonds, respectively.
+        Creat Angles, Dihedrals, Impropers and assing corresponding
+        AngleType, DihedralTypes, and ImproperTypes.
+
+        Parameters
```
bc118, line 312: The backend,  combining_rule, and debug parameters are missing from the doctstring here.

foyer/general_forcefield.py
```diff
@@ -0,0 +1,539 @@
...
+        ----------
+        top : gmso.Topology
+            gmso.Topology object that needed to be parametrized
+        typemap : dict
+            typemap generated by the atomtyper
+        references_file : str, optional, defaut=None
```
bc118, line 318: Spelling/grammer suggestion: "Name of file where force field refrences will be written", where "refrences" is misspelled.

Suggested: "Name of the file where force field references will be written"

foyer/general_forcefield.py
```diff
@@ -0,0 +1,539 @@
...
+            If True, Foyer will exit if parameters are not found for all
+            system bonds.
+        assert_angle_params : bool, optional, default=True
+            If True, Foyer will exit if parameters are not found for all
+            system angles.
+        assert_dihedral_params : bool, optional, default=True
+            If True, Foyer will exit if parameters are not found for all
+            system dihedrals.
+        assert_improper_params : bool, optional, default=True
+            If True, Foyer will exit if parametes are not found for all
```
bc118, line 331: parameters misspelled here

foyer/general_forcefield.py
```diff
@@ -0,0 +1,539 @@
...
+            # Not implemented yet
+            typemap = find_atomtypes(top, forcefield=self)
+        else:
+            typemap = find_atomtypes(top, forcefield=self)
+
+        return typemap
+
+    def _parametrize(self, top=None,
+                        typemap=dict(),
+                        references_file=None,
```
bc118, line 296: I like how we are leaving it open to not exit if all the parameters are not found. :) 👍 )

foyer/general_forcefield.py
```diff
@@ -0,0 +1,539 @@
...
+
+    def _check_parameters(self, top,
+                            assert_bond_params=True,
+                            assert_angle_params=True,
+                            assert_dihedral_params=True,
+                            assert_improper_params=True,
+                            debug=False):
+        """Check if the parameters are fulling filled
+
+        Parameteres
```
bc118, line 445: top and debug are missing from the parameter docstrings

bc118, line 445: Parameteres is misspelled.  Should be "Parameters"

bc118:
A general comment.   I didn't see if residues as an attribute or other way to get residues.  This is important for GOMC and bio simulations as well, as I described in some other PRs.  Does this need to be included in this PR or potentially handled in another PR ?

bc118:
When I look for the bond, angle, and dihedral constant values via the below, I do not see them in the gmso version.

_Maybe I'm missing something, but when I run:_
mol2 = mb.load('c1ccccc1', smiles=True)
omm_oplsaa = foyer.Forcefield(name='oplsaa') # original forcefield object 
gmso_oplsaa = Forcefield(name='oplsaa', strict=False) # new forcefield object 
gmso_top = gmso_oplsaa.apply(mol2, assert_improper_params=False)
pmd_struc = omm_oplsaa.apply(mol2)

print('gmso_top.bonds = ' + str(gmso_top.dihedrals))
print('pmd_struc.bonds = ' + str(pmd_struc.residues))

The gmso version (gmso_top.bonds, gmso_top.angles, gmso_top.dihedrals), provides the atoms in the bond, angle, dihedral, but I don't get the constants like in parmed object.  Can you please elaborate on the differences and if it is accounted for in other code locations....

daico007:
For GMSO, those parameters are stored in the `AtomType` or `BondType` object. So check `gmso_top.bonds[0].bond_type` and the parameters should be there if that `bond_type` exist. similarly, we will have `angle_type` and `dihedral_type` for angle and diehdral

bc118 (changes requested):
The PR looks pretty good from my review.  It takes a little getting used to from parmed structure, but I guess that is expected.  

**_My comments/suggestions before approval:_**
- Most of my comments are minor (spelling ...), minor action required.
- There was 1 question where it looks like it was typemap() the same regardless of the if/else statement.  Just check and confirm this is correct and not an error.  
- The other bigger question or concern was how to handle residues in this section of code or PR... Handle it now, in another PR... ?

bc118:
Maybe I'm missing some info or doing something wrong, but how do I convert a box (mb.Compound) to a GMSO topology?  The below example just seems to turn it into an empty GMSO topology.  Also, see the attached file (to upload the ipynb extension was changed to .txt):

[Example_BC.txt](https://github.com/mosdef-hub/foyer/files/6362010/Example_BC.txt)



Note: I need to keep the box residue name here along the way for GOMC, so is this possible at the moment?.  

**### Input code:**
The error says it basically need to be a from gmso.Topology.
# We can building a box with some SMILES strings/structures
molecule_2 = mb.load('c1ccccc1', smiles=True)
molecule_2.name = 'HEX'
molecule_1 = mb.load('CCO', smiles=True)
molecule_1.name = 'ETO'
box_2 = mb.fill_box(compound=[molecule_1, molecule_2],
                                    density= 200 ,
                                    compound_ratio = [0.8, 0.2] ,
                                    box=[3.0, 3.0, 3.0] )

omm_oplsaa = foyer.Forcefield(name='oplsaa') # original forcefield object
gmso_oplsaa = Forcefield(name='oplsaa', strict=False) # new forcefield object

gmso_box_2 = gmso.Topology(box_2)

_**The last command yeilds the output:**_  
gmso_box_top_dict = ()

foyer/general_forcefield.py
daico007: This is a feature that I need to implement in the future (when we flesh out the `Subtopology` class designs or come up with a better tags system). This is just a place holder right now, so having it True or False does not make any different.

foyer/atomtyper.py
```diff
@@ -1,20 +1,20 @@
...
 
-
 def find_atomtypes(structure, forcefield, max_iter=10):
     """Determine atomtypes for all atoms.
 
     Parameters
     ----------
-    topology : parmed.Structure or TopologyGraph
-        The topology that we are trying to atomtype.
+    topology : parmed.Structure, or gmso.Topology, or TopologyGraph
```
justinGilmer, line 16: The parameter `structure` and `topology` do not match the naming convention.

foyer/atomtyper.py
```diff
@@ -1,20 +1,20 @@
...
+        The topology that we are trying to atomtype, must matched with backend.
```
justinGilmer, line 17: This does not tell the user what backend options are available. This should be explained a bit more.

foyer/general_forcefield.py
```diff
@@ -0,0 +1,507 @@
...
+    Parameters
+    ----------
+    forcefield_files : list of str, optional, default=None
+        List of forcefield files to load.
+    name : str, optional, None
+        Name of a forcefield to load that is packaged within foyer.
+    backend : str, optional, default='gmso'
+        Name of the backend used to store all the Types' information.
+        At this point, 'gmso' is the only valid backend, but this set up
+        allow future backend to be implemented easier.
```
justinGilmer, line 58:
```suggestion
        allows future backends to be implemented easier.
```

foyer/general_forcefield.py
```diff
@@ -0,0 +1,507 @@
...
+        # Preprocessed the input files
+        preprocessed_files = preprocess_forcefield_files(
+                                all_files_to_load,
+                                backend=backend)
+
+        # Load in an internal forcefield object depends on given backend
+        if backend == 'gmso':
+            self._parse_gmso(preprocessed_files, **kwargs)
+        else:
+            raise FoyerError('Backend not supported')
```
justinGilmer, line 98: we should add a test that passes a nonsense backend type and make sure it fails.

foyer/general_forcefield.py
```diff
@@ -0,0 +1,507 @@
...
+              assert_dihedral_params=True, assert_improper_params=True,
+              combining_rule='geometric', debug=False, backend='gmso',
+              *args, **kwargs):
+        """Apply the force field to a molecular topology
+
+        Parameters
+        ----------
+        top : gmso.Topology, or mb.Compound
+            Molecular Topology to apply the force field to
+        references_file : str, optional, defaut=None
```
justinGilmer, line 164:
```suggestion
        references_file : str, optional, default=None
```

foyer/general_forcefield.py
```diff
@@ -0,0 +1,507 @@
...
+            Name of the backend used to store all the Types' information.
+            At this point, 'gmso' is the only valid backend, but this set up
+            allow future backend to be implemented easier.
+
+        Returns
+        -------
+        top : gmso.Topology
+            A parametrized Topology.
+        """
+        # Generate missing connections (angles, dihedrals, and impropers)
```
justinGilmer, line 302:
```suggestion
        # Generate missing connection objects (angles, dihedrals, and impropers)
```

justinGilmer (changes requested):
Some small changes, but other than that, it LGTM!

justinGilmer:
Tested on some oplsaa systems and the parameters were as expected!

justinGilmer (approved):
LGTM!

foyer/general_forcefield.py
```diff
@@ -0,0 +1,502 @@
...
+        if references_file:
+            atom_types = set(site.atom_type for site in top.sites)
+            self._write_references_to_file(atom_types, references_file)
+        return top
+
+    def _parametrize_gmsoFF(self, top, typemap, combining_rule):
+        """Parametrize a Topology with gmso.ForceField"""
+        # Assign AtomTypes
+        for atom in top.sites:
+            atom.atom_type = deepcopy(self.ff.atom_types.get(
```
umesh-timalsina, line 325: Why not use the `get_potential` function here as well?

foyer/general_forcefield.py
```diff
@@ -0,0 +1,502 @@
...
+                                typemap[top.get_index(atom)]['atomtype']))
+        if not all([a.atom_type for a in top.sites]):
```
umesh-timalsina, line 327:
```suggestion
        if not all(a.atom_type for a in top.sites):
```

umesh-timalsina: It might be nice to add this function to Topology class itself. Something like 
```py
def is_fully_typed("atoms"):
```
Just a suggestion.

umesh-timalsina (changes requested):

foyer/general_forcefield.py
```diff
@@ -0,0 +1,502 @@
...
+            tempfile = NamedTemporaryFile(suffix=suffix, delete=False)
+            try:
+                from_foyer_xml(foyer_xml=str(file),
+                          gmso_xml=str(tempfile.name), overwrite=True)
+                tmp_processed_files.append(tempfile.name)
+            except:
+                warnings.warn(f'Could not convert {str(file)}, attempt to read in as is.')
+                tmp_processed_files.append(str(file))
+    else:
+        raise FoyerError('Backend not supported')
```
CalCraven, line 44: Can add a bit more information using the backend variable for this error.

foyer/general_forcefield.py
```diff
@@ -0,0 +1,502 @@
...
+        # Preprocessed the input files
+        preprocessed_files = preprocess_forcefield_files(
+                                all_files_to_load,
+                                backend=backend)
+
+        # Load in an internal forcefield object depends on given backend
+        if backend == 'gmso':
+            self._parse_gmso(preprocessed_files, **kwargs)
+        else:
+            raise FoyerError('Backend not supported')
```
CalCraven, line 103: Same comment as above about making this error a little clearer.

foyer/general_forcefield.py
```diff
@@ -0,0 +1,502 @@
...
+                    break
+
+
+    def _check_parameters(self, top,
+                            assert_bond_params=True,
+                            assert_angle_params=True,
+                            assert_dihedral_params=True,
+                            assert_improper_params=True,
+                            debug=False):
+        """Check if the parameters are fulling filled
```
CalCraven, line 401: """Check if the parameters are filled

foyer/general_forcefield.py
```diff
@@ -0,0 +1,502 @@
...
+    forcefield file of Foyer XML forcefield file
+
+    Parameters
+    ----------
+    forcefield_files : list of str, optional, default=None
+        List of forcefield files to load.
+    name : str, optional, None
+        Name of a forcefield to load that is packaged within foyer.
+    backend : str, optional, default='gmso'
+        Name of the backend used to store all the Types' information.
```
CalCraven, line 60: Seems like theres a small wording issue with the Types'. Not sure if it should be:
'Types'
or 
Type

foyer/general_forcefield.py
```diff
@@ -0,0 +1,502 @@
...
+        At this point, 'gmso' is the only valid backend, but this set up
+        allow future backend to be implemented easier.
```
CalCraven, line 62: will allow future backend to be implemented more easily.

foyer/general_forcefield.py
```diff
@@ -0,0 +1,502 @@
...
+            combining_rule=combining_rule,
+            *args, **kwargs)
+
+    def _run_atomtyping(self, top, use_residue_map=True, **kwargs):
+        """Atomtype the topology
+
+        Parameters
+        ----------
+        top : gmso.Topology
+            Molecular Topology that need to be atomtyped.
```
CalCraven, line 226: Molecular Topology to be atomtyped.

foyer/general_forcefield.py
```diff
@@ -0,0 +1,502 @@
...
+                        assert_angle_params=True,
+                        assert_dihedral_params=True,
+                        assert_improper_params=True,
+                        combining_rule='geometric',
+                        debug=False,
+                        **kwargs):
+        """Parametrize the Topology from the typemap provided
+
+        Assign AtomTypes and BondTypes to Atoms and Bonds, respectively.
+        Creat Angles, Dihedrals, Impropers and assing corresponding
```
CalCraven, line 261: Create Angles, Dihedrals, Impropers and assigning corresponding

CalCraven:
I was playing around with the gmso objects that are created through this conversion. It seems like the charge attributes for each `site` in `topology.sites` is defaulting to 0 even if the site has been typed properly. The site charge can be found by looking at `site.atom_type.parameters["q"]`. If we still want the charge information for that site to be available at the site.charge level, then this charge needs to be properly translated from the atom_type parameter.

foyer/smarts_graph.py
```diff
@@ -352,6 +361,75 @@ def _find_chordless_cycles(bond_graph, max_cycle_size):
     return cycles
 
 
+def _construct_topological_graph(topology):
```
umesh-timalsina, line 364: Is this function necessary now?

foyer/smarts_graph.py
```diff
@@ -2,7 +2,9 @@
 import itertools
 from collections import OrderedDict, defaultdict
 
+import gmso
 import networkx as nx
+import parmed as pmd
```
umesh-timalsina, line 7: Unused Import

foyer/smarts_graph.py
```diff
@@ -2,7 +2,9 @@
 import itertools
 from collections import OrderedDict, defaultdict
 
+import gmso
```
umesh-timalsina, line 5: Unused import?

foyer/smarts_graph.py
```diff
@@ -184,6 +186,13 @@ def _atom_id_matches(atom_id, atom, bond_partners, typemap):
     def find_matches(self, topology_graph, typemap):
         """Return sets of atoms that match this SMARTS pattern in a topology.
 
+        Parameters
+        ----------
+        topology : parmed.Structure or gmso.Topology
```
umesh-timalsina, line 191: This should be `foyer.TopologyGraph`.

foyer/xml_writer.py
```diff
@@ -57,15 +59,20 @@ def write_foyer(self, filename, forcefield=None, unique=True):
...
+            _write_bonds(root, self.bonds, unique)
+        if len(self.angles) > 0 and self.angles[0].type is not None:
+            _write_angles(root, self.angles, unique)
+        if len(self.dihedrals) > 0 and self.dihedrals[0].type is not None:
+            _write_periodic_torsions(root, self.dihedrals, unique)
+        if len(self.rb_torsions) > 0 and self.rb_torsions[0].type is not None:
+            _write_rb_torsions(root, self.rb_torsions, unique)
+
+    # TO DO
+    elif isinstance(self, gmso.Topology):
```
umesh-timalsina, line 74: Why not fail or provide a warning here instead of simply using `pass`?

foyer/general_forcefield.py
```diff
@@ -0,0 +1,564 @@
...
+            raise FoyerError(
+                "Attempting to atom-type using a forcefield "
+                "with no atom type definitions."
+            )
+
+        if self.backend == "gmso":
+            if not isinstance(top, gmso.Topology):
+                mb = import_("mbuild")
+                tmp_top = mb.load(top)
+                top = gmso.external.from_mbuild(tmp_top)
```
umesh-timalsina, line 225: What kind of object/files are supported by `mbuild.load`? It might be nice to do the check?

umesh-timalsina (changes requested):

foyer/general_forcefield.py
```diff
@@ -0,0 +1,564 @@
+"""Foyer general Forcefield class and utility method."""
+import collections
+import glob
+import os
+import re
+import warnings
+import xml.etree.ElementTree as ET
```
umesh-timalsina, line 7: Unused import?

foyer/general_forcefield.py
```diff
@@ -0,0 +1,564 @@
+"""Foyer general Forcefield class and utility method."""
+import collections
+import glob
+import os
+import re
```
umesh-timalsina, line 5: Unused import?

umesh-timalsina (approved):
LGTM. We will have future sprints for any other optimization/changes.

foyer/xml_writer.py
```diff
@@ -57,15 +59,23 @@ def write_foyer(self, filename, forcefield=None, unique=True):
...
+        raise Exception(
+            "Currently cannot write Foyer XML from a gmso.Topology. "
+            "This feature will be implemented in future release."
+        )
```
umesh-timalsina, line 78:
```suggestion
        raise FoyerError(
            "Currently, cannot write foyer XML file from a gmso.Topology. "
            "This feature will be implemented in future releases."
        )
```

PR #369, opened 2020-12-11
OpenMM API Changes Fix

umesh-timalsina:
[`OpenMM 7.5.0`](https://github.com/openmm/openmm/releases/tag/7.5.0) was released and has some changes to the  [`_SystemData`](https://github.com/openmm/openmm/blob/b49b82efb5a253a7c891ca084b3370e181de2ea3/wrappers/python/simtk/openmm/app/forcefield.py#L585) api. This PR is an attempt to fix it.

PR #370, opened 2020-12-11
Pin OpenMM to 7.5

daico007:
### PR Summary:
Pin OpenMM to 7.5 to par with the new fix in #369. 

### PR Checklist
------------

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,4 @@
-openmm
+openmm >=7.5
 mbuild
 parmed
 networkx >=2.0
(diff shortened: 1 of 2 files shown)


justinGilmer (approved):

rsdefever:
Should we pin `openmm` to `7.5` or `>=7.5`? I'm just wondering if pinning `==` would be safer since `foyer` is __very__ dependent on `openmm`.

justinGilmer:
I think >= 7.5 is fine, since the api has been relatively stable for openmm. If it breaks in the future we can specify more detailed versions.

umesh-timalsina (approved):

Issue #364, opened 2020-11-06
Error when trying to parameterize `parmed.Structure`


rsdefever:
I apologize in advance if I'm missing something obvious here.

**Bug summary**

If I convert my `mbuild.Compound` to a `parmed.Structure`, and then try to parameterize it, I'm getting the following error:

```
  File "<stdin>", line 1, in <module>
  File "/Users/ryandefever/software/miniconda3/envs/constrain/lib/python3.7/site-packages/foyer/forcefield.py", line 581, in apply
    verbose=verbose, *args, **kwargs)
  File "/Users/ryandefever/software/miniconda3/envs/constrain/lib/python3.7/site-packages/foyer/forcefield.py", line 632, in parametrize_system
    topology, positions = _topology_from_parmed(structure, self.non_element_types)
  File "/Users/ryandefever/software/miniconda3/envs/constrain/lib/python3.7/site-packages/foyer/forcefield.py", line 145, in _topology_from_parmed
    omm_atom = topology.addAtom(name, element, residues[(pmd_atom.residue.name, pmd_atom.residue.idx)])
KeyError: ('CCC', 0)
```

**Code to reproduce the behavior**

```python
import mbuild
import foyer

ff = foyer.forcefields.load_OPLSAA()

mol1 = mbuild.load("CCC", smiles=True)
mol2 = mbuild.load("COC", smiles=True)

mol1.name = "CCC"
mol2.name = "COC"

box = mbuild.fill_box([mol1, mol2], n_compounds=[100,100], density=700)

box_pmd = box.to_parmed(residues=["CCC", "COC"])
box_ff = ff.apply(box_pmd)
```

I need to go `to_parmed` first so I can specify the residues. Otherwise `parmed` thinks there is a single residue in the entire system. See [mbuild 768](https://github.com/mosdef-hub/mbuild/issues/768) for more.

**Software versions**

- Which version of Foyer are you using? 0.7.6
- Which version of Python (`python --version`)? 3.7.8
- Which operating system? MacOS

daico007:
Fixed by #373

PR #373, opened 2020-12-18
Refactor structure from residue

ahy3nz:
### PR Summary:
Addresses https://github.com/mosdef-hub/foyer/issues/364. I think the bug stems from `foyer._structure_from_residue`. When creating a new structure and adding an already-existing parmed atom, a lot of re-indexing happens to that parmed atom and any associated objects (residues, chains, etc). The easiest solution might be to use `copy` to make entirely new atoms. 

A test has been added, but it's essentially the linked-issue

### PR Checklist
------------


foyer/tests/test_forcefield.py
```diff
@@ -8,9 +9,9 @@
 import parmed as pmd
 import pytest
 
-from foyer import Forcefield
+from foyer import Forcefield, forcefields
```
daico007, line 12: Is there a reason why we import `forcefields` here?

daico007 (approved):
LGTM! Thank you for the fix!

justinGilmer (approved):

PR #388, opened 2021-03-10
Add get_available_forcefield_loaders

mattwthompson:
### PR Summary:
This PR adds a `get_available_forcefield_loaders` function, which returns a list of the functions returning `Forcefield` objects as discovered by the plugin system. I believe there is no other way to systematically access these function in the current API, i.e. from inside Foyer and not the package that plugs into it. In the past I have either called entry points I had in short-term memory or tab-completed `foyer.forcefields.load_` in an interpreter.

### PR Checklist
------------

### DIFF
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -4,8 +4,9 @@ import itertools
 import os
 from tempfile import NamedTemporaryFile
 import xml.etree.ElementTree as ET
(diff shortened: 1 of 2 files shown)


justinGilmer (approved):
lgtm!

umesh-timalsina (approved):

PR #390, opened 2021-03-12
Upload Coverage report using bash script

umesh-timalsina:
Uploads coverage report using `codecov.io` bash script.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -56,8 +56,13 @@ stages:
             displayName: Create Conda env, Activate, Install dependencies, Install Branch
           - bash: |
               source activate test-environment
-              python -m pytest -v --cov=foyer --cov-report= --pyargs foyer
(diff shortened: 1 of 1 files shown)


justinGilmer (changes requested):
Reviewed the failing tests on azure, and they all have the same message:

`__main__.py: error: unrecognized arguments: --no-coverage-upload`

The order of the args in this PR are different from the mbuild PR.

mBuild:
```sh
python -m pytest -v --pyargs mbuild --cov=mbuild --cov-report=html --no-coverage-upload
```

foyer:
```sh
python -m pytest -v --cov=foyer --cov-report=html --pyargs foyer --no-coverage-upload
```

I think all that is needed is swapping around the order of some of the arguments.

I think it should be:

```sh
python -m pytest -v --pyargs foyer --cov=foyer --cov-report=html  --no-coverage-upload
```

Issue #391, opened 2021-03-12
Standerize the use of elements using ele


umesh-timalsina:
Is there a reason we are sub-classing Element from `openMM`in some avenues in the code base as well as using parmed's periodic table for inferring elements in some other avenues in the code base? It seems to me that a lot of this can be standardized using `ele`.

daico007:
the parmed and openmm element are from the old implementation since gmso and ele does not exist. and I think we should replace them soon to eventually move away from parmed an openmm

PR #395, opened 2021-03-31
Integrate ele

daico007:
### PR Summary:

Remove parmed element dependency in the atomtyping step. Relate to #391

### PR Checklist
------------


foyer/atomtyper.py
```diff
@@ -48,8 +48,8 @@ def find_atomtypes(structure, forcefield, max_iter=10):
                 system_elements.add(name)
         else:
             atomic_number = atom_data.atomic_number
-            if 0 < atomic_number <= pt.KNOWN_ELEMENTS:
-                element = pt.Element[atomic_number]
+            if 0 < atomic_number <= len(ele.Elements[0]):
```
rsdefever, line 51: Why not something like:

```python
try:
    element = ele.element_from_atomic_number(atomic_number).symbol
except ElementError:
    raise FoyerError(...)
```
```

umesh-timalsina:
While we are at it can we also take care of https://github.com/mosdef-hub/foyer/blob/a3ec0177507d847ec53136d8c49f424ce188f8a1/foyer/forcefield.py#L16-L17 imports here and remove `foyer/element.py`

daico007:
> While we are at it can we also take care of
> 
> https://github.com/mosdef-hub/foyer/blob/a3ec0177507d847ec53136d8c49f424ce188f8a1/foyer/forcefield.py#L16-L17
> 
> imports here and remove `foyer/element.py`

I think the we still need the OpenMM `elem` here, they are used to infer OpenMM Element objects. I don't think the OpenMM forcefield will accept 3rd party Element object.

foyer/atomtyper.py
rsdefever: @daico007 are you planning on changing this? Your test right now assumes there are no "skipped" elements in the list from `ele`. Ele also provides the `ElementError` for exactly this purpose.

daico007: Sorry, I missed this comment. I will address this in the next commit.

justinGilmer (approved):
LGTM!

justinGilmer:
Also, the RTD build has failed. It seems to be missing ele? Although im not entirely sure yet.

daico007:
I added `ele` to the docs build yml, let see if that fixes the issue

PR #410, opened 2021-04-26
Update conda yaml files to be more consistent and bug fix

rsdefever:
### PR Summary:

* As is, `environment.yml` does not work because the `requests` package is missing.
* `environment-win.yml` is really a development environment `environment-dev-win.yml`. This file was renamed.
* There is no windows equivalent of `environment.yml`. This has been added.

Right now, creating an development environment with `environment.yml` cause the `GAFF foyer` tests to error: `AttributeError: module 'foyer.forcefields.forcefields' has no attribute 'load_GAFF'`. ~~At first I thought this was because GAFF foyer is installed before foyer. Wasn't there some issue with this @justinGilmer @umesh-timalsina. However, I tried re-installing GAFF foyer before foyer so now I am less certain what has gone wrong.~~ If possible, I would also like to get that fixed before we merge this PR.

**Edit**:  The problem is related to the installation order of editable installs with entry points. If the `foyer` installation is editable, it appears that `antefoyer` needs to be installed after `foyer`. If the installation is not editable, the tests fail to even run with the following error:

```
ImportError while loading conftest '/Users/ryandefever/repos/mosdef/foyer/foyer/tests/conftest.py'.
_pytest.pathlib.ImportPathMismatchError: ('foyer.tests.conftest', '/Users/ryandefever/software/miniconda3/envs/foyer-dev/lib/python3.8/site-packages/foyer/tests/conftest.py', PosixPath('/Users/ryandefever/repos/mosdef/foyer/foyer/tests/conftest.py'))
```

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -91,8 +91,8 @@ stages:
             displayName: Add Relevent Channels
 
           - script: |
-              sed -i -E 's/python.*$/python='$(python.version)'/' environment-win.yml
-              conda env create -f environment-win.yml
+              sed -i -E 's/python.*$/python='$(python.version)'/' environment-dev-win.yml
+              conda env create -f environment-dev-win.yml
               call activate foyer-dev
               pip install -e .
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -132,7 +132,6 @@ stages:
               pip install -e .
               cd src/antefoyer
               pip install -e .
-              cd ../../
             displayName: Create a new bleeding test environment
 
           - bash: |
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -62,7 +62,6 @@ stages:
               pip install -e .
               cd src/antefoyer
               pip install -e .
-              cd ../../
             displayName: Create Conda env, Activate, Install dependencies, Install Branch
           - bash: |
               source activate foyer-dev
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -60,10 +60,13 @@ stages:
               conda env create -f environment-dev.yml
               source activate foyer-dev
               pip install -e .
+              cd src/antefoyer
+              pip install -e .
+              cd ../../
             displayName: Create Conda env, Activate, Install dependencies, Install Branch
           - bash: |
               source activate foyer-dev
-              python -m pytest -v --cov=foyer --cov-report=html --pyargs foyer 
+              python -m pytest -v --cov=foyer --cov-report=html --pyargs foyer
             displayName: Run Tests
 
           - bash: |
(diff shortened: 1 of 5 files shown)


daico007 (approved):

daico007:
I think you also need to update the `azure-pipelines.yml` to use the new windows-dev yml

environment-dev-win.yml
```diff
@@ -0,0 +1,24 @@
...
+  - pip
+  - pytest
+  - pytest-azurepipelines
+  - pytest-cov
+  - pytest-timeout
+  - python<=3.8
+  - requests
+  - requests-mock
+  - pip:
+    - "--editable=git+https://github.com/rsdefever/GAFF-foyer.git#egg=gafffoyer"
```
justinGilmer, line 23: Once we resolve the issues we found pertaining to #411, we can merge, but im pretty sure we at least found out that we dont need both `antefoyer` and `GAFF-foyer` installed at the same time since they expose the same name for an entry_point.

rsdefever:
This PR is ready for review. All tests are passing except for codecov.

justinGilmer (approved):

azure-pipelines.yml
```diff
@@ -60,10 +60,13 @@ stages:
               conda env create -f environment-dev.yml
               source activate foyer-dev
               pip install -e .
+              cd src/antefoyer
+              pip install -e .
+              cd ../../
```
jennyfothergill, line 65:
```suggestion
```
I don't think this step is necessary. I think each bash step starts in whatever the root dir is, in this case foyer.

azure-pipelines.yml
```diff
@@ -91,11 +94,13 @@ stages:
...
-              sed -i -E 's/python.*$/python='$(python.version)'/' environment-win.yml
-              conda env create -f environment-win.yml
+              sed -i -E 's/python.*$/python='$(python.version)'/' environment-dev-win.yml
+              conda env create -f environment-dev-win.yml
               call activate foyer-dev
               pip install -e .
-
+              cd src/antefoyer
+              pip install -e .
+              cd ../../
```
jennyfothergill, line 103:
```suggestion
```

azure-pipelines.yml
```diff
@@ -126,6 +131,9 @@ stages:
               pip install -e .
               cd ..
               pip install -e .
+              cd src/antefoyer
+              pip install -e .
+              cd ../../
```
jennyfothergill, line 136:
```suggestion
```

jennyfothergill (approved):
LGTM!
 
Aside: Could you explain the difference between gaff-foyer and antefoyer for someone who is confused? 😅 (Me. I'm confused.)

azure-pipelines.yml
justinGilmer: Reviewing it a bit more, i think you are correct. That does seem to be the case, it will default for each `step` `$Build.SourcesDirectory`

Issue #432, opened 2021-06-28
Add combination rule as a force field property


mattwthompson:
**Describe the behavior you would like added to Foyer**
Mixing rules are important properties of force fields. Foyer should encode them as such in the `Forcefield` object model, much like 1-4 scaling factors (https://github.com/mosdef-hub/foyer/issues/397). Not doing so adds ambiguity around the intended scope of data in a force field and also leads to to issues like https://github.com/mosdef-hub/foyer/issues/431 and issues grad students (me) using the wrong mixing rules for years.

Part of the implementation (telling ParmEd which rule to use) and a TODO are [in the codebase](https://github.com/mosdef-hub/foyer/blob/3b9108247025f11b5614f4c48c0c5418b1e9c7ae/foyer/forcefield.py#L853-L856) already. Just setting `structure.combining_rule` is not sufficient, however, as evidenced by #431, because of how 1-4 scaling factors are passed from OpenMM to ParmEd.

**Describe the solution you'd like**
Given the complexity with storing a non-Lorentz-Berthelot mixing rule in an `openmm.System`, the simplest approach might be to store the mixing rule as a high-level property. Then, if the mixing rule is set to `geometric`, apply at patch onto the ParmEd object before it's ultimately returned by `Forcefield.apply()`.

**Describe alternatives you've considered**
Downstream developers and users could apply a fix that tidies things up, but IMHO this should really be done by foyer and it does not address the original issue of a force field not knowing its combing rule.

**Additional context**
I use "combination rule," "mixing rule", "combining rule," and possibly other terms interchangeably. This is a bad habit of mine but I think it comes from different people and fields not agreeing on a term, so it's probably not going to be fixed anytime soon.

justinGilmer:
Currently satisfied by #433 . Closing for now.

PR #433, opened 2021-06-28
Add Forcefield.combining_rule

### LINKED ISSUE #431: Sigma parameters in 1-4 scaling factors are always computing using Lorentz-Berthlot mixing rules
**Bug summary**

This one goes a little deep, so buckle up. Since Foyer's force field class does not explicitly encode the mixing rule/combination rule, it's somewhat ambiguous how it is ultimately applied. If the force field is meant to encode Lorentz-Berthelot mixing rule, things should work well because that is ParmEd's default and OpenMM assumes that throughout using `openmm.NonbondedForce`. If using geometric mixing rules, I believe the normal flow is to call `Forcefield.apply()` on something and then use the setter of the resulting ParmEd `Structure`, i.e. `my_struct.combining_rule == "geometric"`. Then, ParmEd's writers will _generally_ handle what needs to be handled. For example, writing out to a GROMACS topology file will set `comb-rule` appropriately (3 is geometric):

```
;   File foyer.top  was generated
;   By user: mwt (501)
;   On host: Clausius.local
;   At date: Fri. June  5 14:57:26 2021
;
;   This is a standalone topology file
;
;   Created by:
;   ParmEd:       test_foyer.py, VERSION 3.4.1
;   Executable:   test_foyer.py
;   Library dir:  /Users/mwt/miniconda3/envs/openff-system/share/gromacs/top
;   Command line:
;     test_foyer.py
;

[ defaults ]
; nbfunc        comb-rule       gen-pairs       fudgeLJ fudgeQQ
1               3               no              1            0.5
```

This tells GROMACS to use geometric mixing rules when computing the cross-interactions for each atom type. A deep subtlety here is that the cross-interactions of 1-4 pairs _have already been computed by ParmEd_ based on the values given to it from the `openmm.System` in the `pmd.openmm.load_topology()` call [here](https://github.com/mosdef-hub/foyer/blob/3b9108247025f11b5614f4c48c0c5418b1e9c7ae/foyer/forcefield.py#L832). To repeat, `openmm.NonbondedForce` constructs everything assuming Lorentz-Berthelot mixing rules and encoding geometric mixing rules takes a little more work (https://github.com/openmm/openmm/issues/1859, or @ahy3nz has a deep dive that covers this point: https://ahy3nz.github.io/posts/2019/30/openmm2/). Included in an `openmm.System` is the exceptions (OpenMM language), also knows as adjusts (ParmEd/Amber? language) or pairs (GROMACS language). So the sigma values in the  ParmEd `Structure.adjusts` list have been pre-computed incorrectly.

I think I've reproduced this below, but the resulting errors are small (0.02 kj/mol/molecule) that I could easily be making a mistake that invalidates my conclusions.

I don't think fixing this elegantly would be possible; the easiest way to stop the bleeding would probably involve adding behavior around [here](https://github.com/mosdef-hub/foyer/blob/3b9108247025f11b5614f4c48c0c5418b1e9c7ae/foyer/forcefield.py#L838) that, if using a geometric mixing rule, goes through each `adjust.type.sigma` in `structure.adjusts` and overwrites the existing value with a freshly computed geometric mean. A long-term fix involves a lot of OpenMM hacking that I don't think ParmEd is designed to adjust; I can expand on that if desired but it's no small task. Explicitly storing the combing rule in Foyer's force field object would be necessary for any sustainable solution IMHO.

**Code to reproduce the behavior**

In:

```python3
from typing import TYPE_CHECKING

import foyer  # type: ignore
import mbuild as mb  # type: ignore

if TYPE_CHECKING:
    from parmed import Structure

def _patch_14_with_geometric(struct: "Structure") -> "Structure":
    """Modify sigma values in Structure.adjusts to follow geometric mixing rules"""
    for adj in struct.adjusts:
        sig1 = adj.atom1.sigma
        sig2 = adj.atom2.sigma
        sig_geo = (sig1 * sig2) ** 0.5
        adj.type.sigma = sig_geo

    return struct

oplsaa = foyer.forcefields.load_OPLSAA()
benzene = mb.load("c1ccccc1", smiles=True)

out = oplsaa.apply(benzene)

assert out.combining_rule == "geometric"

# Extract a C-H 1-4 pair from the Structure.adjusts list
for adj in out.adjusts:
    if adj.atom1.name == "C" and adj.atom2.name == "H":
        break

found_14_sigma = adj.type.sigma
sigma_if_lorentz = (adj.atom1.sigma + adj.atom2.sigma) * 0.5
sigma_if_geometric = (adj.atom1.sigma * adj.atom2.sigma) ** 0.5

print(f"{found_14_sigma=}")
print(f"{sigma_if_lorentz=}")
print(f"{sigma_if_geometric=}")
if found_14_sigma != sigma_if_lorentz:
    print("Found 1-4 sigma is not Lorentz-Berthelot!")
if found_14_sigma != sigma_if_geometric:
    print("Found 1-4 sigma is not geometric!")

print("\nApplying hack to rewrite 1-4 interactions with geometric ...")

mod = _patch_14_with_geometric(out)

for adj in mod.adjusts:
    if adj.atom1.name == "C" and adj.atom2.name == "H":
        break

new_14_sigma = adj.type.sigma

if new_14_sigma == sigma_if_lorentz:
    print("Looks like it's still broken \U0001f62C")
elif new_14_sigma == sigma_if_geometric:
    print("Looks fixed! \U0001f60C")
else:
    print("Something is super broken")
```
Out:

```shell
found_14_sigma=2.985
sigma_if_lorentz=2.985
sigma_if_geometric=2.93104077078433
Found 1-4 sigma is not geometric!

Applying hack to rewrite 1-4 interactions with geometric ...
Looks fixed! 😌
```
**Software versions**

- Which version of Foyer are you using? (`python -c "import foyer; print(foyer.version)"`) 0.9.1
- Which version of Python (`python --version`)? 3.8.10
- Which operating system? macOS 11.4

mattwthompson:
### PR Summary:
Fixes #431 and implements #432

Below this line are old comments I think are resolved

----
Right now I can't get the `@property` working, even though I basically copied the existing name and version properties. A script like this should run, but doesn't work for me at the moment locally:

```python3
import foyer
oplsaa = foyer.Forcefield(name="oplsaa")
assert oplsaa.combining_rule == "geometric"
```

As discussed in #432, there are a few names that could be used; I have zero preference and will gladly update it to what the Foyer maintainers prefer.

I updated `foyer.xml_writer._infer_lj_14scale` to assume geometric mixing rules, just to get tests passing. This should probably be updated to take the combining rule as an argument, however.

`Forcefield.apply()` takes an argument `combining_rule`. This is IMHO an anti-pattern, I believe they are properties of force fields and not the typing or parametrization processes. For now I've coded in an exception for when each are defined but not matching; some more care should probably be taken to not break users' workflows. (Though I'd argue there are some cases in which the application of a force field is wrong and workflows _should_ be broken.) The case of using Foyers' OPLS-AA and calling `Forcefield.apply(..., combining_rule="geometric")` should be fine, since the parameters match.

### PR Checklist
------------
- [x] Naming
- [x] Fix loading behavior
- [x] Fix `_infer_lj_14scale`
- [ ] Update test files with proper 1-4 sigmas?
- [x] Tests force fields using each (supported) combination rule

### DIFF
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -639,7 +639,7 @@ class Forcefield(app.ForceField):
                 warnings.warn(
                     "No combining rule found in force field XML file."
                 )
-                return None
+                return "lorentz"
 
     def _create_element(self, element, mass):
         if not isinstance(element, elem.Element):
@@ -714,7 +714,7 @@ class Forcefield(app.ForceField):
         assert_angle_params=True,
         assert_dihedral_params=True,
         assert_improper_params=False,
-        combining_rule="geometric",
+        combining_rule="lorentz",
         verbose=False,
         *args,
         **kwargs,
@@ -837,7 +837,7 @@ class Forcefield(app.ForceField):
         assert_angle_params=True,
         assert_dihedral_params=True,
         assert_improper_params=False,
-        combining_rule="geometric",
+        combining_rule="lorentz",
         verbose=False,
         *args,
         **kwargs,
@@ -876,11 +876,12 @@ class Forcefield(app.ForceField):
 
         if combining_rule != self.combining_rule:
             # TODO: Maybe don't raise an exception here?
-            raise Exception(
+            warnings.warn(
                 f"Combining rule found in force field ({self.combining_rule}) "
                 f"and passed to Forcefield.apply() ({combining_rule}) do not "
-                "match."
+                "match. Using the rule specified in the force field"
             )
+            combining_rule = self.combining_rule
 
         try:
             structure.combining_rule = self.combining_rule
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -874,19 +874,24 @@ class Forcefield(app.ForceField):
             atom_types = set(atom.type for atom in structure.atoms)
             self._write_references_to_file(atom_types, references_file)
 
-        # TODO: Check against the name of the force field and/or store
-        # combining rule directly in XML, i.e.
-        # if self.name == 'oplsaa':
+        if combining_rule != self.combining_rule:
+            # TODO: Maybe don't raise an exception here?
+            raise Exception(
+                f"Combining rule found in force field ({self.combining_rule}) "
+                f"and passed to Forcefield.apply() ({combining_rule}) do not "
+                "match."
+            )
+
         try:
-            structure.combining_rule = combining_rule
+            structure.combining_rule = self.combining_rule
         except ValueError as e:
             # Parmed raises ValueError: combining_rule must be 'lorentz' or 'geometric'
             if str(e).startswith("combining_rule must be"):
                 raise UnimplementedCombinationRuleError(
-                    f"Combination rule {combining_rule} is not implemented"
+                    f"Combination rule {self.combining_rule} is not implemented"
                 )
 
-        if combining_rule == "geometric":
+        if self.combining_rule == "geometric":
             self._patch_parmed_adjusts(structure, combining_rule=combining_rule)
 
         total_charge = sum([atom.charge for atom in structure.atoms])
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -524,25 +524,25 @@ class Forcefield(app.ForceField):
         if validation:
             for ff_file_name in preprocessed_files:
                 Validator(ff_file_name, debug)
-        try:
-            super(Forcefield, self).__init__(*preprocessed_files)
-        finally:
-            for ff_file_name in preprocessed_files:
-                os.remove(ff_file_name)
+        super(Forcefield, self).__init__(*preprocessed_files)
 
-        if isinstance(forcefield_files, str):
-            self._version = self._parse_version_number(forcefield_files)
-            self._name = self._parse_name(forcefield_files)
-            self._combining_rule = self._parse_combining_rule(forcefield_files)
-
-        elif isinstance(forcefield_files, list):
+        if len(preprocessed_files) == 1:
+            self._version = self._parse_version_number(preprocessed_files[0])
+            self._name = self._parse_name(preprocessed_files[0])
+            self._combining_rule = self._parse_combining_rule(
+                preprocessed_files[0]
+            )
+        elif len(preprocessed_files) > 1:
             self._version = [
-                self._parse_version_number(f) for f in forcefield_files
+                self._parse_version_number(f) for f in preprocessed_files
             ]
-            self._name = [self._parse_name(f) for f in forcefield_files]
+            self._name = [self._parse_name(f) for f in preprocessed_files]
             self._combining_rule = [
-                self._parse_combining_rule(f) for f in forcefield_files
+                self._parse_combining_rule(f) for f in preprocessed_files
             ]
+
+        for fp in preprocessed_files:
+            os.remove(fp)
 
         self.parser = smarts.SMARTS(self.non_element_types)
         self._system_data = None
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -38,6 +38,7 @@ from foyer.exceptions import (
     FoyerError,
     MissingForceError,
     MissingParametersError,
+    UnimplementedCombinationRuleError,
 )
 from foyer.utils.external import get_ref
 from foyer.utils.io import has_mbuild, import_
@@ -501,6 +502,7 @@ class Forcefield(app.ForceField):
         self.non_element_types = dict()
         self._version = None
         self._name = None
+        self._combining_rule = None
 
         all_files_to_load = []
         if forcefield_files is not None:
@@ -531,11 +533,16 @@ class Forcefield(app.ForceField):
         if isinstance(forcefield_files, str):
             self._version = self._parse_version_number(forcefield_files)
             self._name = self._parse_name(forcefield_files)
+            self._combining_rule = self._parse_combining_rule(forcefield_files)
+
         elif isinstance(forcefield_files, list):
             self._version = [
                 self._parse_version_number(f) for f in forcefield_files
             ]
             self._name = [self._parse_name(f) for f in forcefield_files]
+            self._combining_rule = [
+                self._parse_combining_rule(f) for f in forcefield_files
+            ]
 
         self.parser = smarts.SMARTS(self.non_element_types)
         self._system_data = None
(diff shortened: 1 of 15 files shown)


justinGilmer:
For testing, what do you think is an amenable solution: should we have two different top files, for each supported combining rule? Or just test the one based on the combining rule specified in the XML?

mattwthompson:
I think both cases should be tested, but the Lorentz-Berthelot case should probably not use OPLS-AA.

The XML writing failures earlier made me a little weary of how much processing is done of the final parametrized structure. This is effectively "backwards" of the normal flow, so some more consideration might be needed there and in similar cases.

mattwthompson:
I'm baffled by the coverage report; it reports that the exception I added is not covered, but the code that raises is it covered. Function definitions `_patch_parmed_adjusts` and `_parse_combining_rule` are reported as not covered, yet their contents _are_ covered.

mattwthompson:
Okay, I've added tests for each case using one-off force field files. Reading the combining rule from the file and setting it on the force field does not work for some reason I can't figure out; I can set a debug around [here](https://github.com/mosdef-hub/foyer/pull/433/files#diff-2e32e78335ccab485367397b205f395d5793e27101a8e63226d79081eb4f2356R543-R545) and find that the data is being parsed correctly, but it's not set on the result. Maybe something is up with how the entry points are configured? **I could please use some help here, it's essential behavior to getting the tests working.** If that's resolved, I should be able to turn around tests and get this ready for review quickly.

justinGilmer:
Quick q, does it work in the instance when it is not a list of forcefield files? Or does it not apply the combination rule in all instances?

justinGilmer:
From my testing, something has to be going out of scope. This is a MWE that fails for all 3 top level properties.

```python3
import foyer
oplsaa = foyer.Forcefield(name="oplsaa")
print('mwe')
print(oplsaa.combining_rule)
print(oplsaa.version)
print(oplsaa.name)
assert oplsaa.combining_rule == "geometric"
```

`name`, `version`, and `combining_rule` seem to be None.

IF HOWEVER,
 you load the XML by using `Forcefield(forcefield_files="path/to/xml")` that does seem to work properly.

justinGilmer:
Can also confirm that version, name, combining_rule all persist if you create a forcefield object using

```python3
import foyer
oplsaa = foyer.forcefields.load_OPLSAA()
print(oplsaa.combining_rule)
print(oplsaa.name)
print(oplsaa.version)
>>> geometric
>>> OPLS-AA
>>> 0.0.1
```

mattwthompson:
I can reproduce that, but I definitely can't explain it on first glance. It's safe to assume that 100% of my usage is from a single file, and I'm not sure how much of Foyer's other uses involve multiple XMLs. This is a pattern inherited from OpenMM (the OpenFF Toolkit inherited as well) that's often used to load in i.e. a water force field and protein force field as separate files.

This is a pretty significant problem for tests and behavior - I can work around it for OPLS-AA by using a common pytest fixture, but loading in a one-off file and inspecting its attributes is pretty important, especially if the combining rule in the force field should line up with `.apply()`.

There was [a test](https://github.com/mosdef-hub/foyer/blob/0aa9683fda92fa05367a748ead6a91312e5a13da/foyer/tests/test_gmso_forcefield.py#L492) for loading in directly from a file, but not one for the `name=""` pattern, which is more commonly-used.

ahy3nz:
Um I'm a little rusty on foyer's forcefield-creation logic, but it seems the forcefield constructor is called _twice_ in a nested fashion? And I think on the second, nested init-logic, that's when the discrepancy in handling `name='oplsaa'` vs `forcefield_files=[files]` happens

ahy3nz:
https://gist.github.com/ahy3nz/4a14cb401067b3ce06ff71be7b89fa06

I don't want to push to this PR or make my own branch, so here's a snippet of what I tried in forcefield.py to get justin's 3 examples to work. 

- use `preprocessed_files` for the metadata parsing, not `forcefield_files`  (also the version, name, combinig rule parsing might need some tweaking to ensure it's a string and not a list)
- I think there should be a tweak to how `os.remove` gets called on xml files. When you pass the `name` flag, the /tmp/ff.xml file is created, but then actually deleted before any of the metadata (version, name, etc) can get processed. When you pass the `ff_files` flag, there's still a persistent ffxml to pull metadata from (or something along those lines)

mattwthompson:
I'm very confused, so I opened a new issue: https://github.com/mosdef-hub/foyer/issues/434

Some of the above ideas work for different ways of loading files (entry points, directly pointing to the file, or using `name`) but not all. I could be doing something wrong, but I do think it's worth splitting it out.

mattwthompson:
Everything's passing locally and I've covered the behavior and tests I'm after. I think if tests pass this is ready for some eyes.

foyer/forcefield.py
```diff
@@ -690,7 +714,7 @@ def apply(
         assert_angle_params=True,
         assert_dihedral_params=True,
         assert_improper_params=False,
-        combining_rule="geometric",
+        combining_rule="lorentz",
```
justinGilmer, line 717: I probably am just forgetting something, but why are we going back to `lorentz` as default?

justinGilmer (approved):
Just a small question, then I think this is g2g when the tests pass.

justinGilmer:
@mattwthompson are all tests passing locally for you? The AZP tests seem to run into a few problematic tests.

foyer/forcefield.py
mattwthompson: This could go either way; my thinking is that it's a more reasonable default value to have as it's more common. If I was presented a force field and forced to assume a combining rule, it's what I'd go with.

If the force field has geometric specified, that overrides the argument here. And `oplsaa.xml` has it, so it should be using geometric.

justinGilmer: Makes sense, so that works for me! Thanks!

mattwthompson:
Everything was passing locally, but I was missing a couple of deps. Most should be fixed, but it also could take a couple of iterations.

mattwthompson:
Okay, passing now!

justinGilmer (approved):
LGTM!

daico007 (approved):
LGTM!

mattwthompson:
Thanks for the quick turnaround @justinGilmer @daico007!

Issue #434, opened 2021-06-29
Metadata parsing issues


mattwthompson:
**Bug summary**

Metadata is not consistently being parsed when force field files are loaded. See [this](https://github.com/mosdef-hub/foyer/pull/433#issuecomment-870143044) thread and some comments before and after it. This blocks #433 and any future metadata additions.

**Code to reproduce the behavior**

In:
```python3
import foyer

oplsaa = foyer.forcefields.load_OPLSAA()
try:
    assert oplsaa.name == "OPLS-AA"
except AssertionError:
    print("load_OPLSAA() did not grab metadata.")

oplsaa = foyer.Forcefield(name="oplsaa")
try:
    assert oplsaa.name == "OPLS-AA"
except AssertionError:
    print("Forcefield(name=\"oplsaa\") did not grab metadata.")
oplsaa = foyer.Forcefield(forcefield_files="foyer/forcefields/xml/oplsaa.xml")
try:
    assert oplsaa.name == "OPLS-AA"
except AssertionError:
    print("Forcefield(forcefield_files=...) did not grab metadata.")
```

Out:
```
/Users/mwt/software/foyer/foyer/validator.py:165: ValidationWarning: You have empty smart definition(s)
  warn("You have empty smart definition(s)", ValidationWarning)
Forcefield(name="oplsaa") did not grab metadata.
```

**Software versions**

- Which version of Foyer are you using? (`python -c "import foyer; print(foyer.version)"`) 0.9.1
- Which version of Python (`python --version`)? 3.9.5
- Which operating system? macOS 11.4

ahy3nz:
https://github.com/mosdef-hub/foyer/blob/3b9108247025f11b5614f4c48c0c5418b1e9c7ae/foyer/forcefield.py#L531-L538

This logic is dependent on the `forcefield_files` arg, but I think it should use `all_files_to_load` variable, since it always gets populated in this code. In the `name=oplsaa` case, the `forcefield_files` var is never populated, but the `all_files_to_load` var DOES get populated. In the other two cases, the `forcefield_files` var gets populated, but so does the `all_files_to_load` var.

If you replace those lines with `all_files_to_load` (`:530,538s/forcefield_files/all_files_to_load/g`), the metadata will be parsed in all 3 cases, but all will return `List[str]` and not `str`. I think changing those 7 lines a possible, quick-fix solution to this metadata parsing issue

PR #435, opened 2021-06-29
Include detailed tests for forcefield metadata

justinGilmer:
### PR Summary:
Previously, the tests in test_forcefield.py were assumed to test that
forcefield metadata like `version`, `name`, etc. were properly loaded
into the `Forcefield` object.

This is only partially true, as there are cases where the files either
go out of scope before that information is gathered, etc. This PR
includes additional tests to ensure that this information is not being
lost prematurely. Big thanks to: @mattwthompson, @ahy3nz for discovering
this issue, creating MWE's for some of these cases, and taking my
investigation further and most likely pinpointing the issue.

### PR Checklist
------------

Refer to issues/PRs: #433 #434 

Note: This does not have @ahy3nz 's fix incorporated yet, just the test cases that fail and currently work.

### DIFF
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -525,7 +525,6 @@ class Forcefield(app.ForceField):
         super(Forcefield, self).__init__(*preprocessed_files)
 
         if len(preprocessed_files) == 1:
-            print("fouind str")
             self._version = self._parse_version_number(preprocessed_files[0])
             self._name = self._parse_name(preprocessed_files[0])
         elif len(preprocessed_files) > 1:
@@ -534,6 +533,8 @@ class Forcefield(app.ForceField):
             ]
             self._name = [self._parse_name(f) for f in preprocessed_files]
 
+        for fp in preprocessed_files:
+            os.remove(fp)
         self.parser = smarts.SMARTS(self.non_element_types)
         self._system_data = None
 
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -522,20 +522,17 @@ class Forcefield(app.ForceField):
         if validation:
             for ff_file_name in preprocessed_files:
                 Validator(ff_file_name, debug)
-        try:
-            super(Forcefield, self).__init__(*preprocessed_files)
-        finally:
-            for ff_file_name in preprocessed_files:
-                os.remove(ff_file_name)
+        super(Forcefield, self).__init__(*preprocessed_files)
 
-        if isinstance(forcefield_files, str):
-            self._version = self._parse_version_number(forcefield_files)
-            self._name = self._parse_name(forcefield_files)
-        elif isinstance(forcefield_files, list):
+        if len(preprocessed_files) == 1:
+            print("fouind str")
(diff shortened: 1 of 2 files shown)


mattwthompson:
Looks like this reproduces the issue (`test_load_metadata_from_internal_name`, the one using `Forcefield(name="")` is failing).

I took Alex's suggested patch (https://github.com/mosdef-hub/foyer/issues/434#issuecomment-870754409) and made a couple fixes so that tests would pass. I haven't thought much about unintended side effects yet.

Here is the commit; feel free to move it around, copy, etc. : https://github.com/mosdef-hub/foyer/commit/22855d422f5770f596356e3cb0d48b49a899f3b9

justinGilmer:
Yeah we essentially had the same design process and handling of the strings. Once this passes and we have the approvals, we can merge. Thanks @mattwthompson  and @ahy3nz for both of your contributions!

mattwthompson:
Does it matter that you're using `preprocessed_files` and I'm using `all_files_to_load`? Both seem to pass tests ... it's probably better to use the pre-processed ones in principle?

mattwthompson:
(Sorry for the misclick 😬)

justinGilmer:
> Does it matter that you're using `preprocessed_files` and I'm using `all_files_to_load`? Both seem to pass tests ... it's probably better to use the pre-processed ones in principle?

That was my thought, we already strip out/handle things that might not behave the way we expect once we have pre-processed.

mattwthompson (approved):
LGTM iff tests pass.

There should be some added confidence if another piece of metadata can be added (#433) without re-engineering this machinery.

ahy3nz:
>>Does it matter that you're using preprocessed_files and I'm using all_files_to_load? Both seem to pass tests ... it's probably better to use the pre-processed ones in principle?

> That was my thought, we already strip out/handle things that might not behave the way we expect once we have pre-processed.

While I agree the pre-processed XML files are safer to try reading, we might have to consider this block of code that is being deleted in this PR. Originally, IIRC @justinGilmer ran into some problems with `/tmp` blowing up when running a lot of atomtyping, so we needed to add this code chunk to remove old XML files. If this chunk of code is removed, this problem may come up again.

https://github.com/mosdef-hub/foyer/commit/748bdd8646342e7a253ee3d3c9944eaad6b83cc9
```python3
        try:
            super(Forcefield, self).__init__(*preprocessed_files)
        finally:
            for ff_file_name in preprocessed_files:
                os.remove(ff_file_name)
```

justinGilmer:
I did add a small section at the end of this init method to iterate through the open files and os.remove them. It runs into the problem where I cant really put a try catch finally around the entire part of that code to always ensure that the files are deleted, but it should handle most cases.

https://github.com/mosdef-hub/foyer/pull/435/files#diff-2e32e78335ccab485367397b205f395d5793e27101a8e63226d79081eb4f2356R536-R537

ahy3nz:
My mistake! Good catch

ahy3nz (approved):

daico007 (approved):
LGTM!

Issue #441, opened 2021-07-07
Changing how foyer handle `combining_rules` when loading in multiple forcefield 


daico007:
**A summary of the question or discussion topic.**
Stemming from this bug #440 here. Currently, when loading in multiple forcefield, their combining rules are stored in a list and then pass into other downstream methods. The list itself can cause an error when it get passed to parmed Structure during the parametrization step. The fix for this bug would be to collapsing the `combining_rules` list to become a string before proceeding to the later steps. However, this is only doable if all the combining rule is the same (either `"lorentz"' or '"geometric"'). I think the more fundamental discussion we should have is to what extent should we allow for the combination of multiple forcefields? I agree with @mphoward in #440, that we the combining should only be possible if the forcefields have the same combining rules(I will have a PR open going for that direction), but I really need everybody's thoughts on this topic.  
@justinGilmer @umesh-timalsina @mosdef-hub/mosdef-contributors

bc118:
In my opinion, if the mixing rules do not match foyer should not infer anything and should fail to build the system.  In general, using different mixing rules should likely be frowned upon, with minor exceptions (TIP3P's standard FF, which is the exact same LB vs geometric mixing, where H-H epsilon is zero).  For this TIP3P water or similar cases, it may make more sense to have two separate XML files for LB and geometric mixing. However, there can be an option to override this.  

If non-standard mixing rules are applied, they need to be done via an NBFIX (Example: CHARMM FF) and/or manually setting the specific pair potentials in the simulation engine control file (Example: LAMMPS).  The best example is in LAMMPS using a hybrid pair potentials with the same mixing rules.  Despite it being the same mixing rules, LAMMPS will still make you specify all the pair potential constants that are not the same (i.e., Ca to O).  Therefore, I believe that there is no reason to make any assumptions on mixing or custom pair potential constants and make the user change them manually or via a FF XML.

We need an option to read custom pair potentials from an XML file into the foyer and write out parameter and control files.  This is currently the major hurdle in utilizing MoSDeF for the CHARMM FF>.

daico007:
That is the current implementation I have over #442 (failed upon merging FF with different combining rule), so we are in agreement in that matter. I don't think we have a way to handle non-standard mixing rules yet, but I guess that can be implemented in the future as need arise.

Issue #440, opened 2021-07-02
Combining two forcefield XML files creates error


mphoward:
**Bug summary**

After updating to foyer 0.9.2, I tried to apply two forcefield files within the same `mbuild.save` command and got this error:

```
/home/mphoward/software/anaconda3/envs/mosdef/lib/python3.9/site-packages/foyer/forcefield.py:879: UserWarning: Combining rule found in force field (['geometric', 'geometric']) and passed to Forcefield.apply() (lorentz) do not match. Using the rule specified in the force field
  warnings.warn(
Traceback (most recent call last):
  File "/home/mphoward/software/anaconda3/envs/mosdef/lib/python3.9/site-packages/foyer/forcefield.py", line 887, in parametrize_system
    structure.combining_rule = self.combining_rule
  File "/home/mphoward/software/anaconda3/envs/mosdef/lib/python3.9/site-packages/parmed/structure.py", line 1571, in combining_rule
    raise ValueError("combining_rule must be 'lorentz' or 'geometric'")
ValueError: combining_rule must be 'lorentz' or 'geometric'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/mphoward/Documents/projects/hydrogel_diff/scripts/mosdef/meoh/../model.py", line 543, in <module>
    make(args.outf, box, crosslinks=crosslinks, compounds=compounds)
  File "/home/mphoward/Documents/projects/hydrogel_diff/scripts/mosdef/meoh/../model.py", line 438, in make
    s.save(
  File "/home/mphoward/software/anaconda3/envs/mosdef/lib/python3.9/site-packages/mbuild/compound.py", line 2029, in save
    conversion.save(
  File "/home/mphoward/software/anaconda3/envs/mosdef/lib/python3.9/site-packages/mbuild/conversion.py", line 982, in save
    structure = ff.apply(structure, **foyer_kwargs)
  File "/home/mphoward/software/anaconda3/envs/mosdef/lib/python3.9/site-packages/foyer/forcefield.py", line 776, in apply
    return self.parametrize_system(
  File "/home/mphoward/software/anaconda3/envs/mosdef/lib/python3.9/site-packages/foyer/forcefield.py", line 891, in parametrize_system
    raise UnimplementedCombinationRuleError(
foyer.exceptions.UnimplementedCombinationRuleError: Combination rule ['geometric', 'geometric'] is not implemented
```

[...]

PR #442, opened 2021-07-07
Collapsing combining of Forcefield when loading from multiple files

daico007:
### PR Summary:
Relate to #440 and #441. Enforcing that combined forcefield (loaded from multiple files) must have the same combining rule. Then collapse the combining rule to a single string (fix the bug found in #440). Also, remove the `combining_rule` option in the `apply` step. 
### PR Checklist
------------


justinGilmer (approved):
LGTM, i think this will lead to foyer 0.10.0 in my opinion. Since we are making a pretty big change to the foyer apply argument, this will be a breaking change.

PR #447, opened 2021-07-27
Migrate azure pipeline environment creation from conda to mamba

justinGilmer:
### PR Summary:
Previously, all conda environment creations use conda for the
dependency resolution, which is notoriously slower than mamba's
implementation.

This can have a noticeable impact on run times during CI as well as
building docker images. Mamba has replaced conda in our dockerfile as
well as our azurepipelines.yml file.

NOTE: mamba is not used for windows builds, due to previous issues
attempting to use mamba.

### PR Checklist
------------


daico007 (approved):
LGTM!

PR #455, opened 2021-08-27
Pin openmm to 7.5

jennyfothergill:
### PR Summary:
New openmm release causing issue in reproducibility study, so let's pin the version until we can determine what the error is.


daico007 (approved):
LGTM! Will merge when tests pass.

justinGilmer (approved):
Once tests pass, LGTM!

PR #457, opened 2021-08-30
Remove simtk references for openmm 7.6

umesh-timalsina:
This PR removes all the simtk references for openmm imports and should be merged after https://github.com/ParmEd/ParmEd/issues/1185 is resolved. Tests will fail for the timebeing.

PR #460, opened 2021-08-31
Update openmm

jennyfothergill:
### PR Summary:
It seems that the openmm version update only broke the way we import this library, so this updates the imports. 

This matters to me because I am trying to get gpu-compiled hoomd from conda working in the reproducibility study, but the openmm version seems to be tied to an older version of cudatoolkit, which prevents it from working.

### PR Checklist
------------

### DIFF
--- a/docs/docs-env.yml
+++ b/docs/docs-env.yml
@@ -5,7 +5,7 @@ channels:
 dependencies:
   - python=3.7
   - gmso
-  - openmm =7.5
+  - openmm=7.6
   - parmed
   - ele
   - pip:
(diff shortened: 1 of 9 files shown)


jennyfothergill:
Currently the build is failing with this error:
```package mbuild-0.10.5-py_0 requires parmed, but none of the providers can be installed```

chrisjonesBSU:
Is this a duplicate of #457 ?

jennyfothergill:
> Is this a duplicate of #457 ?

Yes, and since #457 tests are passing I will close this

PR #461, opened 2021-09-17
Include bond order in TopologyGraph

justinGilmer:
### PR Summary:
`Foyer` currently does not understand bond order in its `TopologyGraph`
representation and in its SMARTS grammar aside from number of connected
neighbors [C;X4], for example.

This PR adds bond order as an attribute when adding bonds in the
`TopologyGraph`, accepted options currently are:

"1"  : single bonds
"2"  : double bonds
"3"  : triple bonds
"ar" : aromatic bonds
"am" : amide bonds
"un": unknown,
"du": dummy
"nc": not connected

The last 3 bond types ("un", "du", and "nc") are subject to change and
may not be here in the final iteration of this PR.

Support has been added when converting parmed structures, openff
topologies, as well as GMSO topologies.
### PR Checklist
------------

### DIFF
--- a/foyer/topology_graph.py
+++ b/foyer/topology_graph.py
@@ -79,7 +79,7 @@ class TopologyGraph(nx.Graph):
         atom_data = AtomData(index, name, atomic_number, element, **kwargs)
         self.add_node(index, atom_data=atom_data)
 
-    def add_bond(self, atom_1_index, atom_2_index):
+    def add_bond(self, atom_1_index, atom_2_index, bond_type="1"):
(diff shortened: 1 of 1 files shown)


foyer/topology_graph.py
```diff
@@ -183,11 +203,22 @@ def from_openff_topology(cls, openff_topology):
                 element=element,
             )
 
+        bond_type_dict = {
```
umesh-timalsina, line 206: This might be better implemented as an `Enum`.

PR #469, opened 2021-10-12
Update `TopologyGraph.from_openff_topology` for new API

mattwthompson:
### PR Summary:
The `Topology` class in the OpenFF toolkit is undergoing a heavy refactor, including removing some API  points that Foyer uses. The refactor will not be in a release for some months, but we are actively testing downstream packages by installing against a feature branch. The parts of the refactor that are relevant here are unlikely to change significantly, but that's not yet guaranteed.

This PR also makes an attempt to support both APIs.

### PR Checklist
------------
 - [x] Code is (approximately) PEP8 compliant (no longer burned on the dev, thanks to `pre-commit`!)


mattwthompson:
Hey folks,

The new version of the OpenFF Toolkit has been released, so I'm able to run these tests without pulling in RC/beta builds. I'm not sure why CI isn't running - is it because I'm pushing to a fork?

mattwthompson:
Tests appear to be fail-fast and failing because we (OpenFF) no longer package on Python 3.7. I opened #513 to possibly mitigate this.

foyer/topology_graph.py
```diff
@@ -161,38 +177,68 @@ def from_openff_topology(cls, openff_topology):
...
+                top_graph.add_bond(atoms_indices[0], atoms_indices[1])
+
+            return top_graph
+
+        else:
+            from openff.units.elements import SYMBOLS
+
+            for atom in openff_topology.atoms:
+                atom_index = openff_topology.atom_index(atom)
+                element_symbol = SYMBOLS[atomic_number]
```
daico007, line 224:
```suggestion
                element_symbol = SYMBOLS[atom.atomic_number]
```

daico007 (approved):
LGTM! coverage is failing due to the code that trying to support old api, so I am not too worried about it.

CalCraven (approved):
This looks fine to me. We don't have a lot of tests for testing this conversion, things we could add would be testing that the element information is properly converted, and testing that the indices of the new topology graph match. 

On a side note, I'm in favor of changing the `element` argument variable in add_atom to `symbol`.

daico007:
Agree with changing the `element` arg to `symbol`, gonna do a quick commit to change the var name

mattwthompson:
Thanks!

PR #488, opened 2021-12-13
Modify XML parsing order of general forcefield

daico007:
### PR Summary:
Add validation step to perform XML conversion only if foyer XML is provided (else load directly by gmso).

### PR Checklist
------------

### DIFF
--- a/foyer/general_forcefield.py
+++ b/foyer/general_forcefield.py
@@ -5,6 +5,7 @@ import os
 import warnings
 from copy import deepcopy
 from tempfile import NamedTemporaryFile
+import shutil
 
 import gmso
 import mbuild as mb
@@ -45,7 +46,7 @@ def preprocess_forcefield_files(forcefield_files=None, backend="gmso"):
                 warnings.warn(
                     f"Could not convert {str(file)}, attempt to read in as is."
                 )
-                os.popen(f"cp {file} {tempfile.name}")
+                shutil.copyfile(file, tempfile.name)
             tmp_processed_files.append(tempfile.name)
     else:
         raise FoyerError("Backend not supported." 'Supports backend: "gmso".')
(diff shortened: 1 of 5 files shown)


bc118 (approved):
I tested this on GOMC-GMSO-MoSDeF and it seems to have fixed the issue.

foyer/general_forcefield.py
```diff
@@ -34,17 +36,18 @@ def preprocess_forcefield_files(forcefield_files=None, backend="gmso"):
...
             try:
+                Validator(ff_file_name=file, debug=False)
                 from_foyer_xml(
                     foyer_xml=str(file),
                     gmso_xml=str(tempfile.name),
                     overwrite=True,
                 )
-                tmp_processed_files.append(tempfile.name)
-            except Exception:
+            except:
```
CalCraven, line 45: Should this be a ValidationError? That's what it seems like the Validator is setup to raise, see [line](https://github.com/mosdef-hub/foyer/blob/c5802ba3dca63c310a309c6b74e5705e4f7ae83b/foyer/validator.py#L74) in validator.py. Might be better to not raise such a general exception, unless there are other errors that might arise from the validation.

foyer/general_forcefield.py
```diff
@@ -34,17 +36,18 @@ def preprocess_forcefield_files(forcefield_files=None, backend="gmso"):
...
                 warnings.warn(
                     f"Could not convert {str(file)}, attempt to read in as is."
```
CalCraven, line 47: This warning could be more descriptive because I think it's one that might be raised commonly if you're already passing gmso `.xmls`. I think something like:

f"Could not convert {str(file)} from a foyer xml, attempting to read in as gmso xml schema."

would suffice

daico007: I have updated the warning as suggested.

foyer/general_forcefield.py
daico007: The current foyer validator is more to check if an atom type is provided provide a malformed SMARTS, or is missing a nonbonded, etc. Trying to validating a GMSO XML will raise error, but not ValidationError, more like XML parsing error (`This element is not expected`). So I think having a general catch here is ok, since if there is any other issues, it would be raised by the GMSO validator.

Issue #491, opened 2021-12-21
Incorrect handling of partially typed systems in  general_forcefield.py


chrisiacovella:
**Bug summary**

It is not uncommon for a force field to only parameterize a subset of the possible topological constraints.  For example, a force field may only have a subset of the possible quartets of atoms described by a dihedral interaction, or in some cases, may not have any dihedral terms as is often common when modeling a solid structure or in many CG force fields.   

To handle this, in forcefield.py, we added in the ability to suppress errors associated with the detection of untyped topological constraints (ie.., bonds/angles/dihedrals/impropers), by allowing a user to set assert_bond_params/assert_angle_params/etc. to False as required.  

For example, in the "old" foyer (i.e., forcefield.py),  if assert_dihedral_params is set to False:
(1) Foyer will not exit with an error when it encounters an untyped quartet of atoms, but instead show a warning
(2) Foyer will return a topology that contains only the typed dihedrals (or be completely empty if none exist)

Currently, general_forcefield.py does 1 correctly, but not 2; it will still include untyped connections in the respective lists, which will cause the writers to fail when they attempt to loop over them.  In forcefield.py, checking for untyped connections involved comparing the typed topology to the list of all possible connections; in general_forcefield.py, only a single topology object exists, and checking involves searching this topology for untyped.  Fixing this should require a relatively simple extension whereby a topology of only the typed parameters is constructed (providing users the ability to toggle this on/off, as a partially typed topology may still be useful, especially during development of a new force field file). I'm working on a some code now. 

**Code to reproduce the behavior**

To demonstrate this I put together a very minimal example where I basically remove angles from my force field to demonstrate the issue.  For this molecule (SiO4) there are 6 possible triplets (i.e., angles), but  since the force field does not contain any entries for angles, we do not parameterize any of them.  While setting assert_angle_params to False will prevent the atomtyping from exiting with an error, if we query the the number of angles in the topology's list of angles, we see it thinks there are 6.   When calling the lammps writer, it will fail when it tries to parse the angle topology because it still thinks there are 6, but none of them have been parameterized with a type (because they do not exist as part of the force field (see: https://github.com/mosdef-hub/gmso/blob/9d324f0988bce893655bd51e4dbeca27cd2fd9d6/gmso/formats/lammpsdata.py#L295).

The forcefield file is included in the zip attachment.

```python
import mbuild as mb
import gmso
import foyer
from foyer import general_forcefield

[...]

PR #493, opened 2022-01-06
Remove untyped

chrisiacovella:
### PR Summary:
This is a fix to issue #491   https://github.com/mosdef-hub/foyer/issues/491 

In short, many forcefields many not have parameters for all possible connections (e.g., no dihedral terms).  While setting, e.g., assert_dihedral_params = False, will prevent the code from throwing an error during atomtyping, untyped connections still are left in the topology returned by apply, which will causes the writers to crash.  This PR adds  remove_untyped_connections argument, that when set to True, will remove untyped connections if the associated connection has assert params set to False (e.g., untyped dihedrals will be removed if assert_dihedral_params = False and remove_untyped_connections = True) .   

[test_remove_untyped.zip](https://github.com/mosdef-hub/foyer/files/7818644/test_remove_untyped.zip)

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.md
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.md
@@ -21,6 +21,6 @@ Please include a code snippet that can be used to reproduce this bug.
 
 **Software versions**
 
-- Which version of Foyer are you using? (`python -c "import foyer; print(foyer.version)"`)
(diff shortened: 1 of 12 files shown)


justinGilmer (approved):

daico007:
@chrisiacovella I think you forgot to commit all the testing XMLs, which are causing the tests to fail.

daico007 (approved):

Issue #495, opened 2022-02-15
Edge case when scaling factors are 0


daico007:
**Bug summary**
There is a bug when typing and parametrizing with forcefields whose scaling factors (lj14 and coulomb14) are 0. Specifically, when we typed and parametrized a mbuild Compound/parmed Structure using forcefield with scaling factors = 0 (i.e., TRAPPE), the scaling factors are not properly carried over when writing to the final form(e.g., when writing out a `.top` file, we have `gen-pairs` as `yes`, and `fudgeQQ` and `fudgeLJ` as 1 instead of 0). 

**Software versions**

- Which version of Foyer are you using? (`python -c "import foyer; print(foyer.__version__)"`)
- Which version of Python (`python --version`)?
- Which operating system?

daico007:
This might be relate to how parmed structure store the 1-4 interactions in `structure.adjusts` and infer the scaling factor in the writer (or just write down under `pairs` section in `.top`). So when scaling factor is 0, it technically does not create any 1-4 interactions, so empty `structure.adjusts` --> default values are used.

PR #496, opened 2022-02-15
Storing meta data in pmd structure

daico007:
### PR Summary:
Relate to #495. Start storing meta data like combining rule and scaling factors in parametrized parmed structure. Currently utilizing the `_Defaults` class from `parmed.gromacs.gromacstop` so terminology may be GROMACS specific, but the info can be stored and access there. 

### PR Checklist
------------

### DIFF
--- a/foyer/forcefield.py
+++ b/foyer/forcefield.py
@@ -29,6 +29,7 @@ from openmm.app.forcefield import (
     RBTorsionGenerator,
     _convertParameterToNumber,
 )
+from parmed.gromacs.gromacstop import _Defaults
 from pkg_resources import iter_entry_points, resource_filename
 
 import foyer.element as custom_elem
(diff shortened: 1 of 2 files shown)


foyer/forcefield.py
```diff
@@ -794,6 +795,31 @@ def apply(
             **kwargs,
         )
 
+        # Start storing scaling factors and combining rule to parmed structure
+        # Utilizing parmed gromactop's _Default class
+
+        # Note: nb_func = 1 (LJ) or 2 (Buckingham), foyer forcefield only support 1 at the moment
```
CalCraven, line 801: Kind of annoying that they handle it by just setting arbitrary integers for the different non-bonded combining rules.  Although I guess that's more on gromacs than on parmed.

CalCraven (approved):
Works fine on my end!

foyer/forcefield.py
daico007: yeah, ideally, parmed would have a way to store this in their structure, but this default class is specific to their gromacs writer, so everything abide to gromacs annotation.

umesh-timalsina (approved):

PR #503, opened 2022-04-19
Use github actions for CI

umesh-timalsina:
This workflow ditches azure pipelines for good. GHA is much better integrated with github and makes life a bit easier with a lot of opensource actions.


daico007:
GHA Windows build with Py38 still has the same fail as Azure: `ImportError: DLL load failed while importing _openmm: The specified module could not be found.`

daico007:
@umesh-timalsina do you think we need to resolve the issue with the readthedocs build before merging this or is it gonna be a another issue?

umesh-timalsina:
I think this is good to go. LGTM!! RTD is another issue. I think this stems from a bloat in the environment size because of added dependencies. We might have to look into mocking few of those.

PR #506, opened 2022-05-24
Modify opls_validation ".top" impropers

CalCraven:
This PR aims to fix bugs in the opls_validation files. More
specifically, improper parameters that were added to those files
for testing do not using the correct "funct" value to designate
them as periodic impropers. They had previously all been given
the value of "1", which designates them as periodic propers. Parmed
doesn't treat periodic propers any differently than periodic impropers,
except to use a "dihedral.improper" boolean. They are still stored in
structure.dihedrals.

This is relevant when doing conversion from a parmed object to GMSO,
as GMSO has explicit container for impropers. This will be necessary
when parameterizing in GMSO as opposed to ParmEd, and testing that
the parameters are stored in this impropers container in GMSO.

See this link for more information about gromacs topology files.
https://manual.gromacs.org/current/reference-manual/topologies/topology-file-formats.html

* Change funct values from "1" to "4" in opls_validation top files
* Add test to "test_opls.py" to validate the correct number of
    impropers are found

### PR Checklist
------------

### DIFF
--- a/foyer/tests/test_opls.py
+++ b/foyer/tests/test_opls.py
@@ -91,17 +91,39 @@ class TestOPLS(BaseTest):
 
     def test_improper_in_structure(self):
         files_with_impropers = [
-            ('o-xylene', 6), ('nitroethane', 1), ('fluorobenzene', 6), ('N-methylformamide', 2), ('formamide', 2),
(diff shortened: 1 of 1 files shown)


daico007 (approved):
LGTM!

PR #474, opened 2021-11-04
changed the atom classes from mirroring their names to what they are …

dyukovsm:
…in gromacs

### PR Checklist
------------
 - [ ] Includes appropriate unit test(s)
 - [ ] Appropriate docstring(s) are added/updated
 - [ ] Code is (approximately) PEP8 compliant
 - [ ] Issue(s) raised/addressed?

PR #512, opened 2022-08-23
drop general forcefield

daico007:
### PR Summary:
With the new GMSO-centric workflow (foyer will be used purely for atomtyping), it has been decided that general_forcefield is no longer needed. This PR removed the general_forcefield and its related tests.

### PR Checklist
------------


CalCraven (approved):
Sounds good, this will reduce a lot of the testing burden we would need and move it to GMSO, where it should be. We probably need a forcefield-utilities backend loader now in Foyer, but that should be another PR. I can get to it after GMSO is covered.

CalCraven:
Is there anything in the Foyer docs about general_forcefield?

daico007:
I don't think we added any docs for the `general_forcefield` (we should have), but I guess it mean less thing to clean up now

daico007:
now also update the atomclass in the `oplsaa.xml` (since this https://github.com/mosdef-hub/foyer/pull/474 is closed)

PR #513, opened 2022-08-31
Drop Python 3.7

mattwthompson:
### PR Summary:
You might want to consider dropping Python 3.7 as it's [out of date](https://numpy.org/neps/nep-0029-deprecation_policy.html) with much of the rest of the ecosystem. I also took the liberty of ditching a few files that probably aren't used anymore.

### PR Checklist
------------


mattwthompson:
> Currently, it looks like the test is failing on py3.10 build but I am not sure what's causing that still.

Micromamba complained about not finding an OpenMM build on Windows. That sounds plausibly a source of confusion I don't want to deal with here, so I've just dropped it.

daico007 (approved):
LGTM!

PR #530, opened 2023-02-02
Update nbfix to use parmed structure.copy

### LINKED ISSUE #529: deep copy in nbfix doesn't preserve all structure attributes
**Bug summary**

When using nbfix, I expect to successfully apply a forcefield and write the structure to a lammps data file. Instead, I obtain the error:
  File ".....mbuild/mbuild/formats/lammpsdata.py", line 170, in write_lammpsdata
    structure = structure.copy(cls=Structure, split_dihedrals=True)
  File ".....python3.8/site-packages/parmed/structure.py", line 613, in copy
    for l in self.links:

**Software versions**

- Which version of Foyer are you using? (`python -c "import foyer; print(foyer.__version__)"`) 11.3
- Which version of Python (`python --version`)? 3.8
- Which operating system? MacOS 13.2

jaclark5:
### PR Summary:

Resolves #529 

By using deepcopy() in foyer.utils.nbfixes.apply_nbfix, the parmed structure loses the following attributes ['links', 'symmetry', 'defaults']. Without links, there is an error when copying the structure in mbuild.lammpsdata.

  File ".....mbuild/mbuild/formats/lammpsdata.py", line 170, in write_lammpsdata
    structure = structure.copy(cls=Structure, split_dihedrals=True)
  File ".....python3.8/site-packages/parmed/structure.py", line 613, in copy
    for l in self.links:

### PR Checklist
------------
 - [X] Includes appropriate unit test(s)
 - [X] Appropriate docstring(s) are added/updated
 - [X] Code is (approximately) PEP8 compliant
 - [X] Issue(s) raised/addressed?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
     - id: black
       args: [--line-length=80]
 - repo: https://github.com/pycqa/isort
-  rev: 5.11.4
+  rev: 5.12.0
   hooks:
     - id: isort
       name: isort (python)
--- a/foyer/utils/nbfixes.py
+++ b/foyer/utils/nbfixes.py
@@ -1,5 +1,5 @@
 """Support non-bonded fixes for various interactions."""
-from copy import deepcopy
+from parmed import Structure
 
 
 def apply_nbfix(struct, atom_type1, atom_type2, sigma, epsilon):
@@ -23,7 +23,7 @@ def apply_nbfix(struct, atom_type1, atom_type2, sigma, epsilon):
     struct : parmed.structure.Structure
         The input structure with the nbfix applied.
     """
-    struct_copy = deepcopy(struct)
+    struct_copy = struct.copy(cls=Structure, split_dihedrals=True)
 
     atom_types = list({a.atom_type for a in struct_copy.atoms})
     for atom_type in sorted(atom_types, key=lambda a: a.name):


jaclark5:
It seems that there is still failure because the parmed.structure.copy() method doesn't make independent atoms. After looking at that parmed method, it should be a copy... I'm not sure how to resolve this. 
Parmed version: 3.4.4

daico007:
Hmm, I can pull the PR and see what's going with the parmed stuff. The CI tests failing seems to be unrelated to the PR itself (more mamba setup, so probably more with the Github action side, I will just rerun the test in a few hours).

daico007:
I pull this local and do some testing, since the CI issue seems to be a bit more complicated than I anticipated. So the only issue is that arise is the original structure (specifically the `structure.has_NBFIX()`, is somehow also changed so causing one extra step to fail. I suspect some object has been copied by reference (probably not the structure obj itself but probably the atom type object (?)). I will need to try a few things and see if we can avoid this issue. 

(Sorry for not having looked at this earlier, things have kinda been hectic)

daico007:
A bit more digging indicates that the issue has to due with how parmed `structure.copy()` works, although the higher level object (like structure, atoms, etc) is a copy (different space in memory), the finer detail (like `atom_type`) is actually only a copy by reference, so any modification of the `atom_type` object of the copy structure will also affect the original structure (in our case is the `structure.has_NBFIX()`)

daico007:
I opened a PR with parmed (https://github.com/ParmEd/ParmEd/pull/1290) try to address this issue

daico007 (approved):

PR #545, opened 2023-10-06
Do not use 'old' OpenFF Toolkit API

mattwthompson:
### PR Summary:

A change I made to the toolkit (https://github.com/openforcefield/openff-toolkit/pull/1661) involved changing how attributes are looked up, which made this check

```
        uses_old_api = hasattr(Topology(), "_topology_molecules")
```

no longer produce the same behavior as when I wrote it. With 0.14.4, released last night, this line erroneously evaluates to `True`. There would be a few ways to fix this, but I'm proposing to just remove the dual support. We are no longer supporting the "old" (< 0.11) versions of the toolkit. Note, of course, that this change wouldn't affect any released version of Foyer.

I also noticed tests aren't run in parallel - it's a small and straightforward fix to get CI to use the multiple cores that [GHA provides 
](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). I'm lazy.

### PR Checklist
------------

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -50,7 +50,7 @@ jobs:
         run: python -m pip install -e .
 
       - name: Test (OS -> ${{ matrix.os }} / Python -> ${{ matrix.python-version }})
-        run: python -m pytest -v --cov=foyer --cov-report=xml --cov-append --cov-config=setup.cfg --color yes --pyargs foyer
+        run: python -m pytest -v -nauto --cov=foyer --cov-report=xml --cov-append --cov-config=setup.cfg --color yes --pyargs foyer
 
       - name: Upload Coverage Report
         uses: codecov/codecov-action@v2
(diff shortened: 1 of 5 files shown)


daico007 (approved):
LGTM! Do we want to pin the openff-toolkit version also?

mattwthompson:
The only constraint is `openff-toolkit >=0.11` but that's already in there. The quirk of 0.14.4 wasn't an API break so it should work all the way back to 0.11

mattwthompson:
Thanks @daico007! Anything I can do to help a release? (If you have a checklist/process in place I'm happy to do as much of it as you are comfortable dumping off)

daico007:
No problem @mattwthompson. I will put a out a patch release for foyer Monday (I may attempt to fix the current Dockerfile before that but there's a good chance it fill fail again)

PR #549, opened 2023-11-28
Clarify error in applying forcefield

jaclark5:
### PR Summary:

Slight change in error string to provide guidance to a user on how to resolve it.

### PR Checklist
------------
 - [NA] Includes appropriate unit test(s)
 - [NA] Appropriate docstring(s) are added/updated
 - [NA] Code is (approximately) PEP8 compliant
 - [NA] Issue(s) raised/addressed?

### DIFF
--- a/foyer/atomtyper.py
+++ b/foyer/atomtyper.py
@@ -226,7 +226,7 @@ def _resolve_atomtypes(topology_graph, typemap):
             )
         else:
             raise FoyerError(
-                "Found no types for atom {} ({}). Forcefield file does not support it.".format(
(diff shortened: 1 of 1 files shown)


foyer/atomtyper.py
```diff
@@ -226,7 +226,7 @@ def _resolve_atomtypes(topology_graph, typemap):
             )
         else:
             raise FoyerError(
-                "Found no types for atom {} ({}).".format(
+                "Found no types for atom {} ({}). Forcefield file does not support it.".format(
```
CalCraven, line 229: I do like this change with additional clarity. I think we could change this wording to something along the lines of 
```suggestion
                "Found no types for atom numbered {} which is atomic number {}. Forcefield file is missing this atomtype, so try to add SMARTS definitions to account for this atom.".format(
```

PR #571, opened 2024-06-20
add python 3.12 and macOS-13 to test matrix

chrisjonesBSU:
### PR Summary:

mBuild and GMSO have started testing with python 3.12 and both versions of MacOS, we should do that here as well.

### PR Checklist
------------

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -11,18 +11,51 @@ on:
     - cron:  "0 0 * * *"
 
 jobs:
-  test:
+  python-test:
     if: github.event.pull_request.draft == false
-    name: Foyer Tests
+    name: Foyer Tests (python)
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
+
+    defaults:
+      run:
+        shell: bash -l {0}
+
+    steps:
+      - uses: actions/checkout@v3
+        name: Checkout Branch / Pull Request
+
+      - name: Install Mamba (Linux)
+        uses: mamba-org/setup-micromamba@v1
+        with:
+          environment-file: environment-dev.yml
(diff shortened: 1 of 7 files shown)


chrisjonesBSU:
The 3.12 tests won't pass until we change the gmso feedstock to pull 3.12

chrisjonesBSU:
The failing test is caused because `mdtraj` is not reading or adding the bonds from the `ethane_customtype.pdb` we are using in `test_from_mbuild_customtype`.

The behavior is different between mdtraj 1.9.9 and mdtraj 1.10, but I don't think it's a general bug related to reading bonds from pdb files. If I use the `ch3.pdb` we have in mbuild's library, both versions of mdtraj read the bonds as expected.


**Using mdtraj 1.9.9:**

```
>>> import mdtraj as md
>>> ethane = md.load("ethane_customtype.pdb")
>>> ethane.top.n_bonds
7
>>>
```

**Using mdtraj 1.10.0:**

```
>>> import mdtraj as md
>>> ethane = md.load("ethane_customtype.pdb")
>>> ethane.top.n_bonds
0
>>>
```

chrisjonesBSU:
Some tests are failing if they use the latest gmso, which has a bug where hoomd needs to be imported for anything using `gmso.external`

Fix is here: https://github.com/mosdef-hub/gmso/pull/828

daico007:
I feel the amount of CI tests are running out of hand. I gonna reduce the tests matrix down a bit (only do python versions tests on linux and platforms tests on the latest python version).

chrisjonesBSU:
> I feel the amount of CI tests are running out of hand. I gonna reduce the tests matrix down a bit (only do python versions tests on linux and platforms tests on the latest python version).

That's a good idea!

chrisjonesBSU:
> I feel the amount of CI tests are running out of hand. I gonna reduce the tests matrix down a bit (only do python versions tests on linux and platforms tests on the latest python version).

This should be implemented now.

daico007 (approved):
LGTM! We can probably propagate this to mbuild and gmso too

PR #574, opened 2024-07-13
Replace `pkg_resources` with `importlib` and add python 3.13 to test suite, bump numpy to 2.0

chrisjonesBSU:
### PR Summary:

I'm trying to fix some of the deprecation errors we're getting in the mosdef packages. One of them is coming from foyer with the use of `pkg_resources`. 

[pkg_resources](https://setuptools.pypa.io/en/latest/pkg_resources.html) will be removed in python 3.13

Note: This PR is not compatible with python 3.9, so we can't merge it until we want to drop 3.9 support and add 3.13.

### PR Checklist
------------


chrisjonesBSU:
It looks like the entry points package in [python 3.9](https://docs.python.org/3.9/library/importlib.metadata.html#entry-points) is different from python >= 3.10. So, that means we can't support both 3.9 and 3.13 whenever it's released. We can table these changes until we're ready to move on to python >= 3.10.

CalCraven:
I don't think it's necessarily the right call here, but we could opt for using 
`from platform import python_version`
to handle the differentiation from python 3.9-3.13 in this PR. Only if we didn't want to wait until we have to fully drop 3.9 to merge this PR.

chrisjonesBSU:
> I don't think it's necessarily the right call here, but we could opt for using `from platform import python_version` to handle the differentiation from python 3.9-3.13 in this PR. Only if we didn't want to wait until we have to fully drop 3.9 to merge this PR.

I say we just wait. Maybe at the next dev meeting we can start talking about the plan to drop 3.9 and add 3.13, which will be officially released next month I believe.

CalCraven (approved):
LGTM

PR #625, opened 2025-07-30
Add more atom type SMARTS defs, new bond, angle and dihedral parameters to `oplsaa.xml`

chrisjonesBSU:
### PR Summary:

The purpose of the PR is to expand available atom types with definitions, bond, angle and dihedral parameters from the ITP files and parameters provided [here](https://virtualchemistry.org/ff.php) and as discussed in [this paper](https://pubs.acs.org/doi/10.1021/ct200731v)

No parameters (epsilons, sigmas, spring constants, etc) were changed for existing entries in `oplsaa.xml`. Some slight modifications to existing atom type SMARTS definitions were made to handle conflicts arising by adding new definitions. All newly added bond, angle and dihedral parameters are at the bottom of their respective sections of the XML, with a comment block indicating the source.

~45 new molecules were added to the opls validation tests

Note: A few `.top` files in the `opls_validation` set of molecules were changed to both reflect the files provided in the link above, and to be more accurate (from what I can tell).

There are still a couple of things remaining, and a few more atom type definitions to sort through, but I wanted to get the PR out sooner rather than later.

### PR Checklist
------------
 - [x] Includes appropriate unit test(s)
 - [x] Appropriate docstring(s) are added/updated
 - [x] Code is (approximately) PEP8 compliant
 - [x] Issue(s) raised/addressed?

PR #629, opened 2025-08-08
Remove all instances of `&` in SMARTS defs from the latest additions to `oplsaa.xml`

chrisjonesBSU:
### PR Summary:

In https://github.com/mosdef-hub/foyer/pull/625, using the `&` (a1 and e2 (high precedence)) in some atom type definitions seems to break the xml parser in forcefield utilities and is causing the GMSO CI to fail. I removed them in this PR, and it doesn't change the atom typing results for the OPLS validation tests. 

### PR Checklist
------------


chrisjonesBSU:
I also found 1 angle entries from #625 that already existed in `oplsaa.xml`, I removed the new entry, so we keepthe original, which matches with the parameters in the validation files.

